PR #194, opened 2018-05-16
Add "open-in-browser" action to urlview_formaction

foxdevilwild:
Fixes issue #23

The urlview window now behaves as expected when recieving
OP_OPENINBROWSER allowing common user-defined macros to work
in this view.

### DIFF
--- a/src/urlview_formaction.cpp
+++ b/src/urlview_formaction.cpp
@@ -25,6 +25,7 @@ urlview_formaction::~urlview_formaction() {
 void urlview_formaction::process_operation(operation op, bool /* automatic */, std::vector<std::string> * /* args */) {
 	bool hardquit = false;
 	switch (op) {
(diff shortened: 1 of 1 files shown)


foxdevilwild:
The "set browser" command - `OP_INT_SET` to be precise - is handled by `formaction::process_op` and thus doesn't need to be handled here.

Different to what the title of #23 suggests, the "set browser" operation has been working as intended all along (for what I can tell), but the "open-in-browser" command which typically follows has no effect. You can verify this with following steps on an unpatched build:

In ~/.newsboat/config:
`macro E set browser <command> %u ; open` (note the use of "open", as opposed to "open-in-browser")

Then run the macro in url-view - it should work as expected.

I am currently using the patched version and can confirm that the common "set browser"-macros work as intended.

PR #460, opened 2019-03-03
Setup directories before attempting to import feeds

thecodefactory:
Setup directories before importing, otherwise the import indicates success but does not succeed unless re-attempted.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -120,6 +120,10 @@ int Controller::run(const CliArgsParser& args)
 
 	configpaths.process_args(args);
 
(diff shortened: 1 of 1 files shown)


thecodefactory:
Behavior without fix (immediately after initial install):

```
$ newsboat -i feedlist.opml 
Import of feedlist.opml finished.
$ newsboat
Starting newsboat 2.13.0...
Loading configuration...done.
Opening cache...done.
Loading URLs from /home/user/.newsboat/urls...done.
Error: no URLs configured. Please fill the file /home/user/.newsboat/urls with RSS feed URLs or import an OPML file.
```

PR #701, opened 2019-11-30
Port FilterParser and Matcher to Rust

tsipinakis:
My attempt at implementing #631, it's time to submit a PR even as a draft because otherwise I feel it'll never be done.

It's partially implemented, a bit messy code and missing some tests but it works in the most usual cases. I'll try to get this to a merge-able state within the next few weeks.

Any comments/reviews welcome.

TODO List from #631:
- [x] re-implement `FilterParser` in Rust, preferably using nom crate (create rust/libnewsboat/src/filterparser.rs)
- [x] port `FilterParser` tests to Rust
- [x] write Rust tests to check that different filter expressions produce expected syntax trees. That's the price we have to pay for implementing our own parsing
- [x] test [this error condition in `Matcher`](https://coveralls.io/builds/25463158/source?filename=src/matcher.cpp#L185) (test/matcher.cpp)
- [x] add `Matchable` trait to Rust, and port `Matcher` to Rust (create rust/libnewsboat/src/matchable.rs and rust/libnewsboat/src/matcher.rs). `Matcher::matches()` should take a trait object of Matchable
- [x] port `Matcher` tests to Rust
- [x] add FFI bindings for `Matchable`:
    - [x] on C++ side, provide `extern "C"` functions that call `has_attribute` and `get_attribute` (add include/rs_matchable.h and src/rs_matchable.cpp)
    - [x] on Rust side, add `struct ForeignMatchable` that implements trait `Matchable` (add rust/libnewsboat-ffi/src/foreignmatchable.rs), calling the aforementioned two C++ functions
- [x] add FFI bindings for `Matcher` (add rust/libnewsboat-ffi/src/matcher.rs)
- [x] edit C++ `Matcher` to use FFI instead of C++ `FilterParser`
- [ ] get rid of C++ `FilterParser`:
    - [ ] remove filter directory
    - [ ] remove test/filterparser.cpp
    - [ ] edit Makefile so it doesn't include "filter" nor tries to build a library from it
- [ ] edit doc/newsboat.txt to remove mentions of Coco/R

Additional items:
- [ ] write comments for each `unwrap()` explaining why it won't ever panic
- [ ] handle matcher errors (`Err` case in `rs_matcher_parse()`)
- [ ] implement `rs_matcher_destroy()` (it's a no-op right now, leaking memory)
- [ ] make sure we have a bullet-proof test suite for this, and Rust passes it.
- [ ] check that no new warnings were introduced
- [ ] document types of all attributes, and what types each operator expects


der-lyse:
(Commenting on lines is currently not working for me)

rust/libnewsboat/src/matcher.rs:63

    i > *a && i < *b

I reckon the range boundaries should be inclusive, thus:

    *a <= i && i <= *b

rust/libnewsboat/src/filterparser.rs
```diff
@@ -0,0 +1,17 @@
...
+enum Value {
+    Str(String),
+    Int(i32),
+    Range(i32, i32)
+}
+
+#[derive(Debug, Clone)]
+enum Expression {
+    And(Box<Expression>, Box<Expression>),
+    Or(Box<Expression>, Box<Expression>),
```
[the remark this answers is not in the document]
tsipinakis: They are otherwise it becomes a recursive struct (or enum in this case) which rustc _really_ doesn't like.

rust/libnewsboat/src/filterparser.rs
```diff
@@ -11,15 +11,18 @@ pub enum Value {
     Range(i32, i32)
 }
 
+#[derive(Debug, Clone)]
+pub struct Condition {
```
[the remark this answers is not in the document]
tsipinakis: This is so we can treat `Condition` as a separate data type than `Expression`. i.e.`check_condition` in `matcher.rs` should make it clear it has to be a condition.

This is the only way I've found to accomplish this.

rust/libnewsboat/src/filterparser.rs
```diff
@@ -0,0 +1,181 @@
...
+    Equal,
+    NotEqual,
+    RegEqual,
+    NotRegEqual,
+    LessThan,
+    GreaterThan,
+    LessThanOrEqual,
+    GreaterThanOrEqual,
+    Between,
+    Contains,
```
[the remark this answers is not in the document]
tsipinakis: Agreed but I couldn't find a better name for that operation. Even in the current parser it's named `CONTAINS`, any ideas?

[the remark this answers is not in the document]
tsipinakis: Agreed but I couldn't find a better name for that operation. Even in the current parser it's named CONTAINS, any ideas?

rust/libnewsboat/src/filterparser.rs
```diff
@@ -15,3 +21,134 @@ enum Expression {
...
+        ) >>
+        (tag)
+    )
+);
+
+fn quoted_string(input: CompleteStr) -> IResult<CompleteStr, Value>{
+    do_parse!(input,
+        tag!("\"")
+        >> value: escaped_transform!(
+                    is_not!("\\\""),
```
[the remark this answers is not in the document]
tsipinakis: It won't match `"\"` because it's missing another quote as a tag, and it panics (another TODO: Proper error handling).

Demo program for the filterparser to try it out until there are proper tests here: https://gist.github.com/tsipinakis/41ec505a6b1997daa18284698df360c0.

rust/libnewsboat/src/filterparser.rs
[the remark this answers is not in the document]
Minoru: I think "contains" is fine; this operation does check if an attribute *contains* a given word. I'm only commenting on other names, which don't fit this pattern: you can't say "1 equal 4", you say "1 equals 4".

Minoru:
Linker errors are gone now. Most of CI builds now fail because of tests. Rust 1.26 builds fail because of `std` not being imported ([build 1](https://travis-ci.org/newsboat/newsboat/jobs/625776733#L962), [build 2](https://travis-ci.org/newsboat/newsboat/jobs/625776724#L780)).

There are new warnings, too. I'll add a TODO item about removing them (because I'll definitely forget to check for them before merging).

rust/libnewsboat/src/filterparser.rs
```diff
@@ -173,7 +173,13 @@ pub fn parse(input: &str) -> Result<Expression, &str> {
         (parse)
     );
     match result {
-        Ok(expr) => Ok(expr.1),
+        Ok(expr) => {
+            if (expr.0.is_empty()) {
+                Ok(expr.1)
+            } else {
+                Err("Trailing characters")
```
Minoru: Sorry this didn't occur to me earlier, but: are these messages shown to the user? If so, let's use `gettextrs::gettext` here.

(Can't find these messages in `po/newsboat.pot`, so either they aren't shown, or weren't internationalized for the C++ version.)

tsipinakis: This isn't shown currently (still need to figure out how to return errors), and this error doesn't exist in our current code.
This is to avoid cases like `((a = b))` parsing and be consistent with our current parser.

tsipinakis:
> Thank you for keeping pushing this PR forward!

No problem! I'm pretty busy lately but I'm still working on this.

> I'd commit C++ tests separately, because it makes it easier to check that they pass not just for the new implementation, but for the old one as well

Very good idea, I'll do it like this once I get to writing the more expansive tests.

rust/libnewsboat/src/filterparser.rs
```diff
@@ -0,0 +1,287 @@
...
+            do_parse!(
+                opt!(tag!("-")) >>
+                take_while1!(|c| is_digit(c as u8)) >>
+                ()
+            )
+        ) >> (num)
+    ).map(|result| {
+        let tuple = result.1;
+
+        let number: i32 = tuple.0.parse().unwrap();
```
pickfire: Does unwrap provides an error handling here?

rust/libnewsboat/src/filterparser.rs
```diff
@@ -0,0 +1,287 @@
...
+        >> tag!("\"")
+        >> (value)
+    ).map(|result| {
+        (result.0, Value::Str(result.1))
+    })
+}
+
+fn number(input: CompleteStr) -> IResult<CompleteStr, Value> {
+    do_parse!(input,
+        num: recognize!(
```
pickfire: Is it possible to do `num: number` like `range` below https://github.com/newsboat/newsboat/pull/701/files#diff-d5d63191785846ad8b3d3e5c6a00cd08R100?

rust/libnewsboat/src/filterparser.rs
tsipinakis: It doesn't, if it's invalid newsboat crashes with this code. However, the filter right above this should make this impossible as it pretty much guarantees it's a valid number.

The only scenario I can see this failing is if a number is so large is doesn't fit within the i32, I'm not sure what the behaviour of `parse()` is in that case.

rust/libnewsboat/src/filterparser.rs
tsipinakis: This is inside the `number` function, where that filter is being defined, so it's not possible :)

pickfire: Oh

rust/libnewsboat/src/filterparser.rs
pickfire: I think `parse` will break if the number is too large too.

rust/libnewsboat-ffi/src/matcher.rs
```diff
@@ -0,0 +1,86 @@
...
+fn matchable_has_attribute(item: *const c_void, attr: *const c_char) -> bool;
+#[no_mangle]
+fn matchable_get_attribute(item: *const c_void, attr: *const c_char) -> *mut c_char;
+}
+
+struct ForeignMatchable {
+    ptr: *const c_void
+}
+
+impl Matchable for ForeignMatchable {
```
pickfire: Why not use `attribute(&self, attr: &str) -> Option<String>` to replace both `has_attribute` and `get_attribute`?

rust/libnewsboat/src/filterparser.rs
Minoru: Yeah, [it'll return an error](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=83e89e7bc006be95ad17e777f2f87993).

Minoru:
I'll resolve a couple of TODO items in #785, so I removed them from the list in the top post.

@tsipinakis, I plan to submit more PRs and get the C++ test coverage to the state where I'm happy with it. After merging those, I'll rebase this PR and fix any breakage I've caused. I'm completely free these days, so hopefully this all will happen over the next week, or even this weekend.

Finally, when all of that is done, I'd like to take over the rest of the tasks here, and complete this PR in time for 2.19. Would you mind me doing that, @tsipinakis? I don't intend to steal your thunder, just wish this code to be a part of `master` branch already :)

tsipinakis:
> Would you mind me doing that, @tsipinakis? I don't intend to steal your thunder, just wish this code to be a part of master branch already :)

No problem, feel free to take over. I'd also like to see this done and merged one way or the other and given that I hadn't been able to work more than a few hours on this for a while it's probably for the best :)

pickfire:
@tsipinakis What is left for this? Need help?

Minoru:
@pickfire hi there. The top post actually has a TODO list, and in a comment just above yours me and @tsipinakis agreed that I'd take care of this PR. I don't yet see an opportunity to parallelize this.

pickfire:
@Minoru Then any other things related to rust that I can take care of other than this patch?

Minoru:
@pickfire Take a look at the [rust-port label](https://github.com/newsboat/newsboat/labels/rust-port). But they all involve some C++, we don't have anything that's purely in Rust yet.

The process of creating more such issues is constrained by my ability to isolate and refactor some module or subsystem into a state where I can more or less clearly describe what things to port. That's funny, it means that in order to write more Rust, we need more C++ programmers :)

Minoru:
It appears that we can't produce user-friendly error messages with Nom 4. Parsing `"title =¯ \"foo\""` produces the following:

```rust
nom::Err::Error(
    nom::Context::Code(
        CompleteStr("title =¯ \"foo\""),
        nom::ErrorKind::Alt
    ))
```

This tells us that the `alt!` parser failed at the end of the input string. We get *exactly* the same error for `"a = b"` or `"a = \"b"` (save for the text inside `CompleteStr`). There's no way to turn that same error into three different error messages.

Enabling "verbose-errors" feature of Nom changes nothing (a bit surprisingly).

nom's author [claims](http://unhandledexpression.com/general/2019/06/17/nom-5-is-here.html) that Nom 5 is better in this regard (we're still on 4.2.3). I'm investigating that now.

Minoru:
Ported fmtstrformatter to Nom 5. Used the new [`VerboseError`](https://docs.rs/nom/5.1.0/nom/error/struct.VerboseError.html) type during debugging. Still doubtful that we'll be able to provide helpful one-line error messages, but at least there's some hope now.

pickfire:
@Minoru Do you push your partial changes somewhere? I am still not quite used to nom, I hope I can learn more about it by looking at your commits.

Minoru:
Not pushing anywhere yet, because it's just an experiment to see if Nom 5 will help with errors. If it does, I'll create one new PR after 2.19 release (to update fmtstrformatter), and then update this PR to use Nom 5 too.

Minoru:
The situation with Nom 5 is a bit better:

```rust
assert_eq!(
    parser::<VerboseError<&str>>("title =¯ \"foo\""),
    Err(nom::Err::Error(VerboseError {
        errors: vec![
            ("¯ \"foo\"", VerboseErrorKind::Nom(ErrorKind::TakeWhile1)),
            ("¯ \"foo\"", VerboseErrorKind::Nom(ErrorKind::Alt)),
            ("title =¯ \"foo\"", VerboseErrorKind::Nom(ErrorKind::Alt))
        ]
    }))
);
```

We still get the same error tree for radically different inputs, but at least we can now report a position that thwarted the parser.

Minoru:
Just got a question on IRC: a person tried to use `tags =~ tagname` and got an error saying "couldn't parse filter command". All because they didn't wrap the "tagname" in quotes. Perhaps we *do* need those descriptive error messages after all.

Minoru:
Okay, here's the next roadblock: C++ parser stores all values as strings, and `Matcher` prefers to do string comparison. Some operators do convert strings to numbers, but when they do, they only use the numeric *prefix* of the string. There are a few implications of this:

1. Numbers can be used in place of strings and even regular expressions:
    * `guid = 123` will match any article with GUID of "123";
    * `title =~ 007` will match any article that contains "007" in the title.

2. Strings can be used in place of numbers:
    * `age = "7"` will match any article that is seven days old.

3. Equality does string comparison, *regardless of argument type*:
    * `age = 07` won't ever match an article, because numeric attributes don't have leading zeros (which is surprising, since "age" is a number and "07" looks like a number that is obviously equal to seven).

4. Comparison operators convert arguments to numbers, but only use the numeric prefix of the string, and fallback to zero:
    * `age > "7 days"` will work, but " days" part is actually ignored -- in reality, the value of "age" attribute is compared to number seven. This is a very dangerous case, because users might think that e.g. `age > "7 weeks"` or `age > "7 months"` should work too -- and those expressions *will* be parsed, but they'll still match articles that are over 7 days old (because "age" is measured in days);
    * `title >= 0` might seem nonsensical, but it's a valid expression. It will match any article except ones that have negative numeric prefix in their title, e.g. "-60 days left until Python 2 reaches EOL".

I believe that all of these are bugs: things like `age > "7"` might be innocuous, but they're bundled with actual pitfalls like `age > "7 weeks"`. The rest of them are just plain confusing.

What does the ideal design look like to me? On the left-hand side, we can have either a number, or a string. On the right-hand side, we can have a string, a number, or a range. For most operators, LHS and RHS types have to be the same. The types on the left- and right-hand sides of an operator have to (roughly) match. For `between`, LHS has to be a number, and RHS has to be a range.

This PR already adds types for RHS, but LHS is still stringly-typed, and it'd be cumbersome to change at the moment: LHS value is created by C++, so we'll have to teach all implementers of `Matchable` to return either a string or a number, and then drag that over FFI. Doable, but a substantial effort. Luckily, it's enough to know the operator and (sometimes) the RHS type to know what LHS should be converted to.

That's what I'll be implementing, then. Documentation will have to be updated as well (added an item to the TODO list in the top post).

Minoru:
I ended up with a long, convoluted Git history, so I'm going to aggressively rebase stuff to make it more reviewable. This will make my history completely unrelated to this PR, so I'm closing this.

This also means that, most probably, @tsipinakis will be absent from the Git log for this, but I'll credit him in the CHANGELOG. On the bright side, he's also going to be absent from the Git blame… :)

PR #720, opened 2020-01-16
PB Queueloader: Remove PbController dependency

dennisschagt:
This will simplify writing tests for QueueLoader because it:
- allows testing with different queue file location;
- removes the need for setting up PbController with serveral environment variables.

### DIFF
--- a/include/download.h
+++ b/include/download.h
@@ -1,6 +1,7 @@
 #ifndef PODBOAT_DOWNLOAD_H_
 #define PODBOAT_DOWNLOAD_H_
 
+#include <functional>
 #include <string>
 
 namespace podboat {
@@ -21,7 +22,7 @@ class PbController;
 
 class Download {
 public:
-	explicit Download(PbController* c = 0);
+	explicit Download(std::function<void()> cb_require_view_update);
 	~Download();
 	double percents_finished() const;
 	const std::string status_text() const;
@@ -57,7 +58,7 @@ private:
 	double totalsize;
 	double curkbps;
 	unsigned long offs;
-	PbController* ctrl;
+	std::function<void()> cb_require_view_update;
 };
 
 } // namespace podboat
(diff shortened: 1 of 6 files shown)


dennisschagt:
Reacting to https://github.com/newsboat/newsboat/pull/717#issuecomment-574319187
> > I'm planning to write some tests but before that I think some small refactoring is useful to make writing tests easier.
> 
> That's risky: you might inadvertently break something during refactoring, and won't know about it because there are no tests. Writing tests first doesn't eliminate the risk completely, but lowers it (significantly IMHO).
> 
> You can try your approach, but before submitting the PR, please run a simple check: remove all refactorings (with interactive rebase, or by cherry-picking tests onto "master"), and make sure the tests pass.
> 
> It might be useful to look at code coverage while writing tests. You can either submit a PR marked as draft, and view Coveralls reports online (the bot will post a link after CI build); or run `test/generate_coverage_report.sh` locally.

I agree that it is useful to write tests before refactoring but there is a lot of overhead when writing tests for QueueLoader currently (before applying this PR):
- [QueueLoader requires a pointer to a PbController instance](https://github.com/newsboat/newsboat/blob/712a7e7cb9731cd75df7d2918802fb2c27c2cfe4/src/queueloader.cpp#L23-L25) so all QueueLoader tests would have to instantiate PbController
- Quite a lot of things are being executed in the [constructor](https://github.com/newsboat/newsboat/blob/712a7e7cb9731cd75df7d2918802fb2c27c2cfe4/src/pbcontroller.cpp#L129-L183) and [`run`](https://github.com/newsboat/newsboat/blob/712a7e7cb9731cd75df7d2918802fb2c27c2cfe4/src/pbcontroller.cpp#L190-L322) method of PbController. It takes some effort to setup multiple environment variables (e.g. $HOME/$XDG_xx) to make sure the test does not mess with config/queue/db files of people executing the test on their own machine.

I'm confident the changes in this PR don't introduce any issues.
If you really want to, I could look into writing test (without the refactor) but I think in this specific case it is a waste of time.
I much rather write a few tests after these changes are approved/merged in.

PR #868, opened 2020-04-08
Add unread flag field which allows formatting with conditional sequence

dennisschagt:
From irc:
```
07Apr-07:33:15 < _Warl0ck> Is there any way to change the characters for flags? For example if I wanted a fancy unicode character instead of N for unread article. I just customized neomutt like this so I would like to do it here as well
07Apr-07:35:44 < yitz> Yes. You can set key bindings in the config somewhere
07Apr-07:45:10 < _Warl0ck> yitz: Key bindings? I am simply interested in formatting the article list
07Apr-07:48:19 < yitz> Oh. Totally misunderstood.
07Apr-07:48:19 < yitz> https://newsboat.org/releases/2.18/docs/newsboat.html#_format_strings
07Apr-07:48:20 -url:#newsboat- _ The Newsboat RSS Feedreader
07Apr-07:49:21 < yitz> Hrm. Not sure if you can customize what %n gets replaced with
07Apr-07:50:13 < _Warl0ck> yeah that's what I'm trying to do :/
07Apr-07:50:49 < _Warl0ck> pasted a bunch of FontAwesome icons into my neomutt config and it looks amazing
07Apr-07:51:33 < yitz> https://github.com/newsboat/newsboat/blob/09ec32244dca3f5699cb736636c8b5a2105d31ef/src/itemlistformaction.cpp#L1380
07Apr-07:51:34 -url:#newsboat- _ newsboat/itemlistformaction.cpp at 09ec32244dca3f5699cb736636c8b5a2105d31ef _ newsboat/newsboat _ GitHub
07Apr-07:52:01 < yitz> If you are up for building it yourself, it's a one char change
07Apr-07:52:49 < yitz> Man. It's been years since I opened the sourcecode for this project...
07Apr-07:53:03 < _Warl0ck> Nice. Maybe I could make a quick patchfile that substitutes some of the common flags
07Apr-09:07:35 < nikos> Maybe %?n?new-char? would work?
```

The suggested `%?n?new-char?` does not work because the `%n` specifier is not empty when the article is read (it still contains whitespace for alignment).

This PR adds a new `%N` specifier which results in an empty string, allowing use of the conditional sequence.

### DIFF
--- a/src/itemlistformaction.cpp
+++ b/src/itemlistformaction.cpp
@@ -1019,6 +1019,7 @@ std::string ItemListFormAction::item2formatted_line(const ItemPtrPosPair& item,
 	FmtStrFormatter fmt;
 	fmt.register_fmt('i', strprintf::fmt("%u", item.second + 1));
 	fmt.register_fmt('f', gen_flags(item.first));
+	fmt.register_fmt('N', item.first->unread() ? "N" : "");
 	fmt.register_fmt('D',
 		utils::mt_strf_localtime(
 			datetime_format,
--- a/doc/newsboat.asciidoc
+++ b/doc/newsboat.asciidoc
@@ -1002,7 +1002,8 @@ Identifier:Meaning
 [[feedlist-format-i]]<<feedlist-format-i,+i+>>:Feed index
 [[feedlist-format-l]]<<feedlist-format-l,+l+>>:Feed link
 [[feedlist-format-L]]<<feedlist-format-L,+L+>>:Feed RSS URL
-[[feedlist-format-n]]<<feedlist-format-n,+n+>>:"unread" flag field
+[[feedlist-format-n]]<<feedlist-format-n,+n+>>:"unread" flag field ("N" if unread, " " if read)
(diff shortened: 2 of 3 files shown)


tsipinakis:
How about ignoring whitespace in conditionals instead, it's technically 'empty'? IMO I'd expect an all whitespace strings to be considered empty.

src/itemlistformaction.cpp
```diff
@@ -1019,6 +1019,7 @@ std::string ItemListFormAction::item2formatted_line(const ItemPtrPosPair& item,
 	FmtStrFormatter fmt;
 	fmt.register_fmt('i', strprintf::fmt("%u", item.second + 1));
 	fmt.register_fmt('f', gen_flags(item.first));
```
dennisschagt, line 1021: @Minoru: Do you think it is useful to add the other flags (normally part of `f` with [gen_flags()](https://github.com/newsboat/newsboat/blob/09ec32244dca3f5699cb736636c8b5a2105d31ef/src/itemlistformaction.cpp#L1374-L1390)) as well? (i.e. a format specifier for `item->deleted()` and a format specifier for `item->flags().length()`)

dennisschagt:
> How about ignoring whitespace in conditionals instead, it's technically 'empty'? IMO I'd expect an all whitespace string to be considered empty.

Sounds good!

PR #877, opened 2020-04-11
Reloader: Use ScopeMeasure and apply some small refactoring/cleanup

dennisschagt:
I noticed that the `Reloader` class was still using manual calculations with `time()` instead of `ScopeMeasure`.
While fixing that, I noticed some related code which could use some improvement.

### DIFF
--- a/include/reloader.h
+++ b/include/reloader.h
@@ -71,6 +71,7 @@ public:
 		unsigned int size,
 		bool unattended = false);
 
+private:
 	/// \brief Notify in various ways that there are new unread feeds or
 	/// articles.
 	///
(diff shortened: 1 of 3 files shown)


include/reloader.h
```diff
@@ -89,6 +82,15 @@ class Reloader {
 	/// notification will contain \a msg passed.
 	void notify(const std::string& msg);
 
+protected:
```
Minoru: Class members are private by default, so this commit effectively relaxed the access. No classes inherit from `Reloader`, so the point is kind of moot, but still: why change the access level?

dennisschagt: My bad, will fix it and force-push this small change.

dennisschagt:
In case it is useful, the output of `git range-diff` for the force-pushed changes:
image

Minoru:
Thank you!

PR #915, opened 2020-05-01
Use "-C panic=abort" instead of "-Zno-landing-pads"

dennisschagt:
Needed because rust removed support for "-Zno-landing-pads":
https://github.com/rust-lang/rust/pull/70175

This caused our Test Coverage build to fail:
https://travis-ci.com/github/newsboat/newsboat/jobs/326201747#L651-L656


dennisschagt:
I'm not sure why `-Zno-landing-pads` was specified and thus have no idea if there is any problem in dropping it.
I will add an issue instead of this trial-and-error PR approach.

PR #936, opened 2020-05-13
Add wrap scroll feature

Limero:
This will add an optional wrap scroll feature. When enabled, moving down while on the last item in a list will wrap around to the top and vice versa.

There are equivalent features in ranger, lf, etc. https://ranger.github.io/ranger.1.html#wrap_scroll-bool

Note: I'm not sure if you want the Newsboat config added to Podboat as it's currently missing. So I just set the wrap scroll value to `false` for it in `pbview.cpp`.


Limero:
I have added wrapping for page up/down now.

I'm not sure about prev/next though, because wrapping scroll when you can't see what it will move to is confusing, like on the article list.

I don't think there are any more stuff missing after looking through the KeyMap list.

Limero:
I have added it to Podboat now too :smile:

PR #1106, opened 2020-07-22
Support both old and new rust toolchains simultaneously (clippy conversion related)

dennisschagt:


dennisschagt:
@Minoru: I think my latest commit makes it possible to drop https://github.com/newsboat/newsboat/pull/1106/commits/a1cee13d9a3bc29af4fd6d73087f0e2cd1b286c2.
Let me know if you want me to try that.

Also, feel free to make any changes yourself to this branch or just cherry-pick my commit onto [your existing PR](https://github.com/newsboat/newsboat/pull/1100).

dennisschagt:
> I'm in two minds about this. On one hand, this lets people with older toolchains run Clippy locally and see just the warnings that they introduced. On the other hand, we have no way of knowing when to drop the old lint for good, so they'll just keep piling up. Also, anyone who suppresses a Clippy warning might need to implement this trick too, which implies they need to know about it.

Somewhere in the future we might be able to use something like the following instead of those 4 lines: 
```
#[cfg_attr(version("1.45.0"), allow(clippy::useless_conversion))]
#[cfg_attr(not(version("1.45.0")), allow(clippy::identity_conversion))]
```
but it depends on https://github.com/rust-lang/rfcs/pull/2523 which is not yet implemented (https://github.com/rust-lang/rust/issues/64796).

It might also be possible to use [`cfg_attr`](https://doc.rust-lang.org/1.9.0/book/conditional-compilation.html#cfg_attr) by manually adding an attribute/feature when running clippy but I don't really know how/where to do that.

PR #1168, opened 2020-09-02
Refactor ColorManager

dennisschagt:
Improvements:
- Don't handle 0 color rules as a special case (related to `colors_loaded`)
- Make it explicit that there are always the same amount of foreground colors, background colors and attribute vectors
- Remove code duplication between `ColorManager::set_pb_colors()` and `View::apply_colors()`


Minoru (approved):
Looks good, awesome job! :+1:

PR #1175, opened 2020-09-05
Add documentation to open specific URLs in browser

A1RO:
While trying to figure out the command to open a specific URL in the browser, I found that the keybindings `0-9` were not properly documented.

In my case, I wanted to create a macro to send podcast audio URLs to mpv. The commands `one`, `two`, `three`, etc. are exactly what I was looking for (I actually found them in the article view help dialog). Here's an example:

```sh
browser firefox
macro p set browser "mpv"; one ; set browser "firefox"
```

In this PR, I add the missing commands under the "Available Operations" section of the docs.
Please note that i'm not setup to test the end result. Feel free to modify as needed.


Minoru (approved):
Thank you, this looks great! I restarted the failing CI build (looks like a sporadic failure with apt.llvm.org), and will merge this once the build finishes.

PR #1243, opened 2020-10-24
Add contrib/exportOPMLWithTags.py

jartigag:
As discussed in #871, I add this script. It's useful to export the urls file to OPML including tags.


Minoru (approved):
Thank you!

PR #1274, opened 2020-11-10
Continue loading feeds from external services even when `urls` file can't be opened

dennisschagt:
Fixes https://github.com/newsboat/newsboat/issues/1273.

Tested with a local Nextcloud setup to verify that this actually works (I should have done that before submitting https://github.com/newsboat/newsboat/pull/1269).


Minoru (approved):
Thanks! :+1:

PR #1278, opened 2020-11-11
External APIs: Only mark unread items as read

dennisschagt:
@arjan-s at https://github.com/newsboat/newsboat/issues/220#issuecomment-725632690:
> There is one bug in the current code: it also performs the API call for articles that were already marked read. Changing this to only mark unread articles as read would improve the performance a lot.

Good idea, implemented in this PR.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -610,12 +610,14 @@ void Controller::mark_all_read(unsigned int pos)
 	}
 
 	if (feed->is_query_feed()) {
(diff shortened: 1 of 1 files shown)


Minoru (approved):
:+1:

arjan-s:
Awesome! I'll test this later today and let you know.

arjan-s:
Works as advertised. Thank you!

PR #1287, opened 2020-11-14
Nextcloud: Mark multiple items read in single request

dennisschagt:
Supersedes https://github.com/newsboat/newsboat/pull/1283, using @Minoru's idea https://github.com/newsboat/newsboat/pull/1283#issuecomment-726965670:
> I also wonder if it would make more sense to add mark_all_read() to RemoteAPI. We can provide a default implementation that calls mark_read() in a loop, and OcNewsApi can provide an optimized implementation. This neatly avoids the problem with shared state.

### DIFF
--- a/include/ocnewsapi.h
+++ b/include/ocnewsapi.h
@@ -17,6 +17,7 @@ public:
 	std::vector<TaggedFeedUrl> get_subscribed_urls() override;
 	bool mark_all_read(const std::string& feedurl) override;
 	bool mark_article_read(const std::string& guid, bool read) override;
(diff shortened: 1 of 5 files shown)


Minoru (approved):
Looks great, thanks!

I don't have NextCloud News, so just to make sure: you did test this, right? If so, you can just merge this. Otherwise we can ask arjan-s to verify.

dennisschagt:
> I don't have NextCloud News, so just to make sure: you did test this, right?

Yes, I have a small test setup on which I verify my changes in the OcNewsApi.

PR #1289, opened 2020-11-14
Register (and ignore) `run-on-startup` option in podboat

dennisschagt:
Fixes https://github.com/newsboat/newsboat/issues/1288


Minoru (approved):
Excellent, thank you!

PR #1294, opened 2020-11-15
ItemList: Add items to list before changing the position

dennisschagt:
Fixes https://github.com/newsboat/newsboat/issues/1293


Minoru (approved):
Great, thank you very much @dennisschagt!

PR #1306, opened 2020-11-23
Use flushinp() instead of ignoring input events one-by-one

dennisschagt:
A small improvement on top of https://github.com/newsboat/newsboat/pull/1259.

This reduces screen-flickering (caused by drawing the `EmptyFormAction`) and might be a bit faster (single curses call without waiting for 1 ms timeout).

### DIFF
--- a/include/view.h
+++ b/include/view.h
@@ -57,7 +57,7 @@ public:
 		return keys;
 	}
 	void set_tags(const std::vector<std::string>& t);
-	void drop_queued_input(Stfl::Form& form);
+	void drop_queued_input();
 	void pop_current_formaction();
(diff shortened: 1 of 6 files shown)


Minoru (approved):
Nice! I think `form_action` variables can be dropped now, because they aren't used anymore? Just merge this if I'm wrong, or fix it and merge then :)

Thanks for the pull request!

dennisschagt:
> I think form_action variables can be dropped now, because they aren't used anymore

Good point, done: https://github.com/newsboat/newsboat/compare/7f56233d74e6a3596fea4180254de6aec029d944..291035819291a89da710ff58a851a6ec79ca3188

PR #1310, opened 2020-11-24
Various small improvements

dennisschagt:


include/reloader.h
```diff
@@ -41,9 +41,6 @@ class Reloader {
 	/// made through \a easyhandle, unless it's nullptr, in which case
 	/// method creates a temporary handle that is destroyed when method
 	/// completes.
-	// TODO: check that the value passed via "max" is always obtained from
```
Minoru, line 44: Do we have an issue for this? I couldn't find it.

src/opml.cpp
```diff
@@ -107,55 +110,30 @@ void rec_find_rss_outlines(
...
-				if (!found) {
-					LOG(Level::DEBUG,
-						"opml::import: added url = %s",
-						url);
-					urlcfg.get_urls().push_back(
-						std::string(url));
+				auto& urls = urlcfg.get_urls();
+				if (std::find(urls.begin(), urls.end(), quoted_url) == urls.end()) {
+					LOG(Level::DEBUG, "opml::import: added url = %s", quoted_url);
+					urls.push_back(std::string(quoted_url));
```
Minoru, line 124: `quoted_url` is already of required type, so I think the call to `std::string` can be dropped here.

Minoru (approved):
Nice changes!

src/opml.cpp
dennisschagt: Good point, done: https://github.com/newsboat/newsboat/pull/1310/commits/8968822269c33dac06e2a662f97035917693c4a9

include/reloader.h
dennisschagt: The TODO seemed to point to a possible improvement (not a bug).
I implemented it in https://github.com/newsboat/newsboat/pull/1310/commits/84421408b37ac8c96bcb14e836332b9e38cd7b25.

include/reloader.h
```diff
@@ -42,7 +42,7 @@ class Reloader {
 	/// method creates a temporary handle that is destroyed when method
 	/// completes.
 	void reload(unsigned int pos,
-		unsigned int max = 0,
+		bool show_progress = false,
```
Minoru, line 45: It looks like `show_progress` serves the same purpose as `unattended`. I guess that `max == 0` case wasn't actually important — it's inconceivable that a user would reload something if `FeedContainer` has no feeds in it. My suggestion is to drop `show_progress` altogether, and remove `max` from the doc comment.

Minoru (changes requested):

Minoru:
Also, please rebase onto master to avoid problems with Travis.

include/reloader.h
dennisschagt: The `max == 0` case indeed could happen when `FeedContainer` has no feeds in it.
However, there are more opportunities for `max == 0` (or with the current changes, `show_progress == false`):
- https://github.com/newsboat/newsboat/blob/9370e77d476853bb738bdf3579843d9d1975da1d/src/feedlistformaction.cpp#L121
- https://github.com/newsboat/newsboat/blob/9370e77d476853bb738bdf3579843d9d1975da1d/src/itemlistformaction.cpp#L412

I prefer to keep the current changes, but I'm fine with changing it if you want to always show `[<pos>/<num_feeds>]` in the status (currently it is not included when the user executes a single `reload` operation).

I did update the doc comment so it now mentions `show_progress` instead of `max`.

dennisschagt:
> Also, please rebase onto master to avoid problems with Travis.

Done:
image

include/reloader.h
Minoru: Oh, sorry, I misunderstood the logic ._. `show_progress` and `unattended` are indeed different: the former is only responsible for the `<pos>/<num_feeds>` part, while the latter can show/hide the whole message.

The changes to the doc comment look good; no other changes needed. Sorry for wasting your time on this!

Minoru (approved):
Thanks!

include/reloader.h
dennisschagt: No problem, thank you for the review!

PR #1317, opened 2020-11-27
Migrate FsLock rust binding to cxx crate

dennisschagt:
I wanted to get some experience with the cxx crate.

I expect the `pid_t` to `std::int64_t` conversion to be fine:
> From https://man7.org/linux/man-pages/man3/pid_t.3.html:
> pid_t: … According to POSIX, it shall be a signed integer type, and the implementation shall support one or more programming environments where the width of pid_t is no greater than the width of the type long.

This PR depends on https://github.com/newsboat/newsboat/pull/1304.
I will rebase onto master when that PR is merged (and change the target branch).

### DIFF
--- a/include/fslock.h
+++ b/include/fslock.h
@@ -1,30 +1,22 @@
 #ifndef NEWSBOAT_FSLOCK_H_
 #define NEWSBOAT_FSLOCK_H_
 
+#include "fslock.rs.h"
+
 #include <string>
 #include <sys/types.h>
-
-extern "C" {
-
-	void* rs_fslock_new();
-
-	void rs_fslock_free(void* ptr);
-
-	bool rs_fslock_try_lock(void* ptr, const char* new_lock_filepath, pid_t* pid);
-
-} // extern "C"
 
 namespace newsboat {
 
 class FsLock {
 public:
 	FsLock();
-	~FsLock();
+	~FsLock() = default;
 
 	bool try_lock(const std::string& lock_file, pid_t& pid);
 
 private:
-	void* rs_fslock = nullptr;
+	rust::Box<fslock::bridged::FsLock> rs_object;
 };
 
 } // namespace newsboat
--- a/rust/libnewsboat/src/fslock.rs
+++ b/rust/libnewsboat/src/fslock.rs
@@ -97,11 +97,4 @@ impl FsLock {
             false
         }
     }
-
-    pub fn try_lock_ffi(&mut self, new_lock_path: &str, pid: &mut i64) -> bool {
-        let p: &mut libc::pid_t = &mut 0;
-        let result = self.try_lock(Path::new(new_lock_path), p);
-        *pid = i64::from(*p);
-        result
-    }
 }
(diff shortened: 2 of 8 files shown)


include/fslock.h
```diff
@@ -1,30 +1,22 @@
...
-
 namespace newsboat {
 
 class FsLock {
 public:
 	FsLock();
 	~FsLock();
 
-	bool try_lock(const std::string& lock_file, pid_t& pid);
+	bool try_lock(const std::string& lock_file, std::int64_t& pid);
```
Minoru, line 16: I'd keep `pid_t`, and use `libc::pid_t` on the Rust side. A lot of our dependencies depend on libc crate, so we can use it quite confidently — we won't be dropping it anytime soon.

rust/libnewsboat/src/fslock.rs
```diff
@@ -97,4 +97,11 @@ impl FsLock {
             false
         }
     }
+
+    pub fn try_lock_ffi(&mut self, new_lock_path: &str, pid: &mut i64) -> bool {
```
Minoru: I would very much prefer not to add wrappers to the libnewsboat crate: it muddles the API, and we'll have to track them down later when they aren't needed anymore. (The compiler won't help with that because `pub` methods are assumed to be used somewhere outside the crate.)

Instead, I'd write a function in the libnewsboat-ffi crate. It's uglier (one has to call `method(object, arg1, arg2)` instead of `object->method(arg1, args2)`), but I think it's better than polluting the API in libnewsboat crate. I think it's easier to *show* what I mean rather than describe it, so I pushed 2235c8fe into your branch. Feel free to squash it, drop it, or amend it to your liking — it's just a demonstration of what I envision, not a proposal for inclusion into this PR.

include/fslock.h
dennisschagt: Amended into: https://github.com/newsboat/newsboat/pull/1317/commits/682bb8d0011071abfe4aabf8970f3a0887b1f647

rust/libnewsboat/src/fslock.rs
dennisschagt: > I would very much prefer not to add wrappers to the libnewsboat crate

Agreed. I had some trouble getting that to work. Thank you for the example!

include/fslock.h
Minoru: It looks like only C++ side got amended? Rust still uses `i64`. I haven't tried it — is there a problem using `libc::pid_t` with cxx?

dennisschagt: From what I understand, cxx only supports a limited number of types in structs, arguments, and returns:
https://cxx.rs/bindings.html#built-in-bindings-reference
`libc::pid_t` is not in that list.

image

Minoru: Ah. Okay then, let's keep a wrapper on the Rust side.

dtolnay: Supporting stuff like this is in scope for CXX but is just not implemented yet. See https://github.com/dtolnay/cxx/issues/529 for an issue to follow.

Minoru:
Hmm, this auto-closed when I merged #1304. I'm not sure if PRs can be reopened. Oh well. Please submit a new one then :)

PR #1366, opened 2020-12-19
Config parsing: Allow tabs between macro arguments

dennisschagt:


Minoru (approved):
Thanks! This looks good, and I'm ready to merge this before the 2.22 release, but I got one question: why the switch to `multispace` variants? They [accept carriage returns and line feeds](https://docs.rs/nom/6.0.1/nom/character/complete/fn.multispace0.html), which can't occur in the input for `tokenize_operation_sequence()` because that function only ever gets a part of a line. Furthermore, accepting more than we strictly need might turn into a bug if we integrate this parser more tightly with `ConfigParser` in the future. Am I missing some use-case where CR and NL are valid parts of an operation sequence?

dennisschagt:
> but I got one question: why the switch to multispace variants? They accept carriage returns and line feeds, which can't occur in the input for tokenize_operation_sequence() because that function only ever gets a part of a line

Because the carriage return and line feed do not occur in the input and I thought it would be easier to read.

> Furthermore, accepting more than we strictly need might turn into a bug if we integrate this parser more tightly with ConfigParser in the future.

Sounds like a good reason; I changed it so it only accepts spaces and tabs now.

Minoru (approved):
Okay, let's merge this! Thank you for the work.

PR #1370, opened 2020-12-21
Don't call `run()` on empty formactions

dennisschagt:
Fixes https://github.com/newsboat/newsboat/issues/1369.

- [x] ~I believe I got all relevant places but will put a little more time in trying to find other places where we checked for a `nullptr` formaction (before we introduced the `EmptyFormAction`).~ -> Checked all places which call `stfl_run()` using `git grep "\\.run("`. Apart from the 2 changed locations, all of those are only called from the main thread (which is blocked while the editor or browser is running).


Minoru (approved):
Looks correct to me. Thanks for the swift fix!

PR #1376, opened 2020-12-22
Make sure we can't accidentally overwrite a feed's rssurl

dennisschagt:
Previously, a feed's rssurl could get overwritten accidentally, leading to https://github.com/newsboat/newsboat/issues/685.
This PR makes sure we can no longer overwrite a feed's rssurl (by marking it const).

### DIFF
--- a/include/rssfeed.h
+++ b/include/rssfeed.h
@@ -126,11 +126,6 @@ public:
 
 	void update_items(std::vector<std::shared_ptr<RssFeed>> feeds);
 
-	void set_query(const std::string& s)
-	{
-		query = s;
-	}
-
 	bool is_query_feed() const
 	{
 		return rssurl_.substr(0, 6) == "query:";
--- a/include/rssfeed.h
+++ b/include/rssfeed.h
@@ -101,8 +101,7 @@ public:
 	std::shared_ptr<RssItem> get_item_by_guid_unlocked(
 		const std::string& guid);
 
-	/// \brief User-specified feed URL. Can't be empty, otherwise we wouldn't
-	/// be able to fetch the feed.
+	/// \brief User-specified feed URL.
 	const std::string& rssurl() const
 	{
 		return rssurl_;
--- a/include/rssfeed.h
+++ b/include/rssfeed.h
@@ -19,7 +19,7 @@ class Cache;
 
 class RssFeed : public Matchable {
 public:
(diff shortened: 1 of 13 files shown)


Minoru:
Sorry I didn't take a look earlier, looks like this PR fell through the cracks.

I like the idea of correctness-by-construction, but I think this PR falls a bit short of it because it allows empty URLs — failing to uphold a guarantee that `rssurl()` getter promises: https://github.com/newsboat/newsboat/blob/1752592a581f433025c60c49cb2e429b44f33251/include/rssfeed.h#L105-L110

(Yes, the guarantee might've not been upheld before this PR either. All the more reasons to make it right :)

I think an empty URL can be treated as a programming mistake, so the check can be done with `assert()`.

This check implies more changes to the tests, so I didn't look too close at them. This also means you can force-push if you want.

Does that sound reasonable? Or should I have read the tests and discover that my idea won't fly? :)

dennisschagt:
Thank you for the review and sorry for my late reply

> I think an empty URL can be treated as a programming mistake, so the check can be done with assert().

Currently, an empty rssurl signals that a feed contains search results, see for examples:
https://github.com/newsboat/newsboat/blob/f1295894e134e2589fc29c90582efd7b9c0477e7/src/itemlistformaction.cpp#L473-L493
https://github.com/newsboat/newsboat/blob/f1295894e134e2589fc29c90582efd7b9c0477e7/src/itemlistformaction.cpp#L1452-L1465

My idea is to store search results in a separate class.
That makes it possible to require a non-empty rssurl in the regular `RssFeed` type and allows us to remove several unused fields from the (to-be-introduced) `SearchResults` type.
I rather don't do that in this PR.

What do you prefer?
- Remove the un-true claim from the code documentation (`Can't be empty, otherwise we wouldn't be able to fetch the feed`).
- Merge as-is
- Delete this PR (I wouldn't mind)
- ??

Minoru:
Ah, I see. Makes sense! Let's remove the un-true claim then. Meanwhile, I'll take a quick look at the tests here.

Minoru (approved):
Tests look good to me, feel free to merge after removing the outdated claim from the `rssurl()` docstring. And thanks for the pull request!

dennisschagt:
Thanks!

Force pushed:
image

PR #1380, opened 2020-12-24
Code coverage: Ignore generated code

dennisschagt:
Specifically, this ignores code genereated by cxxbridge in `target/cxxbridge/libnewsboat-ffi/src`:
image
image


Minoru (approved):
Yup, makes sense: this is "glue code" that we don't write by hand, so we don't care if it's covered or not.

PR #1382, opened 2020-12-24
Miscellaneous Podboat code improvements/cleanup

dennisschagt:


src/pbcontroller.cpp
```diff
@@ -299,8 +298,8 @@ int PbController::run()
 
 	std::cout << _("done.") << std::endl;
 
-	ql = new QueueLoader(queue_file, cfg,
-		std::bind(&PbController::set_view_update_necessary, this, true));
+	ql.reset(new QueueLoader(queue_file, cfg,
```
Minoru, line 293: Looking at podboat.cpp, it's obvious that `run()` runs for almost the entire lifetime of `PbController`, and the object is not re-used afterwards. Thus, is it possible to make `QueueLoader` an ordinary, non-pointer member of `PbController`?

src/pbview.cpp
```diff
@@ -193,12 +192,9 @@ void PbView::run(bool auto_download, bool wrap_scroll)
 		case OP_PB_DOWNLOAD: {
 			if (ctrl->downloads().size() >= 1) {
 				const auto idx = downloads_list.get_position();
-				if (ctrl->downloads()[idx].status() !=
-					DlStatus::DOWNLOADING) {
-					std::thread t{PodDlThread(
-							&ctrl->downloads()[idx],
-							ctrl->get_cfgcont())};
-					t.detach();
+				auto item = ctrl->downloads()[idx];
```
Minoru, line 195: `auto` drops references, so `item` here is a `Download`, and we're passing a reference to this local variable to the thread in `start_download()`. I think this should be changed to `auto&`.

src/pbcontroller.cpp
```diff
@@ -113,9 +113,7 @@ bool PbController::setup_dirs_xdg(const char* env_home)
 		config_dir + NEWSBEUTER_PATH_SEP + config_file;
 
 	/* in data */
-	cache_file =
-		xdg_data_dir + NEWSBEUTER_PATH_SEP + cache_file;
-	lock_file = cache_file + LOCK_SUFFIX;
+	lock_file = xdg_data_dir + NEWSBEUTER_PATH_SEP + LOCK_SUFFIX;
```
Minoru, line 116: Am I reading this correctly that with XDG, Podboat locks ~/.local/share/newsboat/.lock (assuming XDG_DATA_DIR is is not set or is ~/.local/share), but with dotdirs, it locks ~/.newsboat/pb-lock.pid?

I understand you didn't make this change, I just wonder if I'm missing something here, and you probably read this code a bit more than I did.

Minoru (changes requested):
Looks cool, I'm glad the class becomes smaller and hopefully easier to understand!

src/pbcontroller.cpp
dennisschagt: It looks like that requires some more formatting which I rather do at a later time (in a different PR).
The refactoring would be needed to make the value for `queue_file` available while initializing `ql` in the `PbController` initializer list.

src/pbcontroller.cpp
dennisschagt: > Am I reading this correctly that with XDG, Podboat locks ~/.local/share/newsboat/.lock (assuming XDG_DATA_DIR is is not set or is ~/.local/share), but with dotdirs, it locks ~/.newsboat/pb-lock.pid?

You are right, definitely looks strange although it's not really a bug (the name is not mentioned in the documentation and it properly prevents a second podboat instance from starting).

dennisschagt:
I decided to rebase directly, range-diff:
image

src/pbview.cpp
dennisschagt: Thanks for catching that!
Fixed by amending/rebasing: https://github.com/newsboat/newsboat/pull/1382/commits/4acab0faf99451c73b4101c631ca353c5f79b058#diff-8ba3707b9e587217ca7dfc65695eae819d8c1bd28e10ff720484cd6033dab44aR195

src/pbcontroller.cpp
Minoru: This is bizarre. Just to make sure that it's not a regression, I checked `git blame` and this code was so from the beginning: https://github.com/newsboat/newsboat/commit/404c19172ebeaca01a1aae8250c6820bc334d742 I think I'll go add this to the docs, because I'm sure I'll forget it at some point.

src/pbcontroller.cpp
Minoru: Ah, right, because `queue_file` has to be initialized with XDG/dotfile logic. Got it, thanks for the explanation!

Minoru (approved):
Cool, thanks for the PR! :+1:

PR #1421, opened 2021-01-09
Fix issue related to `save-all` path

dennisschagt:
Fixes https://github.com/newsboat/newsboat/issues/1418

### DIFF
--- a/src/dirbrowserformaction.cpp
+++ b/src/dirbrowserformaction.cpp
@@ -99,17 +99,6 @@ bool DirBrowserFormAction::process_operation(Operation op,
 						fn.push_back(NEWSBEUTER_PATH_SEP);
 					}
 
-					std::string fnstr =
-						f.get("filenametext");
-					std::string::size_type base =
-						fnstr.find_last_of(NEWSBEUTER_PATH_SEP);
-					if (base == std::string::npos) {
-						fn.append(fnstr);
-					} else {
-						fn.append(fnstr,
-							base + 1,
-							std::string::npos);
-					}
 					set_value("filenametext", fn);
 					do_redraw = true;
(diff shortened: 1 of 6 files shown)


src/dirbrowserformaction.cpp
```diff
@@ -295,24 +284,20 @@ void DirBrowserFormAction::init()
...
-
-		dir = save_path;
-	}
+	LOG(Level::DEBUG,
+		"view::dirbrowser: save-path is '%s'",
+		save_path);
 
-	int status = ::chdir(dir.c_str());
-	LOG(Level::DEBUG, "view::dirbrowser: chdir(%s) = %i", dir, status);
+	int status = ::chdir(save_path.c_str());
```
der-lyse, line 293: I reckon this can now be `const` as well.

der-lyse (approved):
Fixing bugs by removing code are the very best fixes! :-)

I confirm, this fixes the bug.

src/dirbrowserformaction.cpp
dennisschagt: Thanks, suggestion squashed into https://github.com/newsboat/newsboat/pull/1421/commits/0f31b9da5dd8ded7a7178c1986867ceb5af802b1.

Minoru (approved):
Thanks! This looks excellent to me. :+1:

The first two commit messages were a bit confusing because at the time, `dir` *was* used in both classes. But the last two commits made it all make sense in the end :)

Anyone can merge this once the CI is done and Lyse gave his approval.

der-lyse (approved):
Reapproved! :-)

dennisschagt:
Thank you for the reviews!

Issue #1491, opened 2021-02-24
Setting toggle syntax broken


tsipinakis:
**Newsboat version (copy the output of `newsboat -v` or the first line of `git show`)**: cd9d777 (also broken in 2.22.1, 2.22)

Config file (copy from ~/.newsboat/config or ~/.config/newsboat/config):
```
macro m set mark-as-read-on-hover!
```

Steps to reproduce the issue:

1. Open any feed

2. Run the provided macro

Instead of toggling `mark-as-read-on-hover` it displays an invalid syntax error.
Bisect points to 0fb0c28fb974a1a4f3030de2350e7bb6bcc0cf6a, which makes sense.

dennisschagt:
I somehow overlooked that feature, thanks for the report!
I'm working on a fix now.

PR #1492, opened 2021-02-24
Re-introduce toggle (!) and reset(&) options for `set` in macros

dennisschagt:
Fixes https://github.com/newsboat/newsboat/issues/1491.

Tested with config:
```
macro m set mark-as-read-on-hover!

macro a set feedlist-title-format "changed title"
macro b set feedlist-title-format&
```

### DIFF
--- a/include/formaction.h
+++ b/include/formaction.h
@@ -45,6 +45,7 @@ public:
 	void recalculate_widget_dimensions();
 
 	virtual void handle_cmdline(const std::string& cmd);
+	bool handle_single_argument_set(std::string argument);
 
 	bool process_op(Operation op,
 		bool automatic = false,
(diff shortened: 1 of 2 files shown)


Minoru (approved):
Thanks!

That's what we get for implementing the same thing in two different places I guess :) It'd be nice to extract this into a new private method, and call that both from here and from `FormAction::handle_cmdline()`.

If it's more work than you're willing to do right now, it's fine: just move my comment into an issue, and merge this PR. Thanks once again for such a prompt fix!

dennisschagt:
Force-pushed after resolving merge conflicts.

> It'd be nice to extract this into a new private method, and call that both from here and from FormAction::handle_cmdline().

Done for the `set <variable>[!&]` variant.
The `set <variable> <value>` variant has multiple differences between the two implementations (like backtick handling) so I did not try to de-duplicate those.

Minoru (approved):
Awesome, thank you very much! :+1:

PR #1499, opened 2021-02-27
Simplify formatting in HelpFormAction

dennisschagt:
The main improvement is the unrolling of a complex loop.
I believe this makes it easier to understand the function.

The changes have been split into several commits.
If useful, I'm completely fine with squashing those commits together.


Minoru (approved):
Very nice cleanup, thanks!

PR #1509, opened 2021-03-05
Use maximum value for Stfl::run() timeout

dennisschagt:
Fixes https://github.com/newsboat/newsboat/issues/563.


Minoru (approved):
This is very pragmatic. I'm ashamed I didn't think of this solution myself :)

Thanks for the pull request! I'll merge it once I figured out what's up with the Coveralls CI job, which currently segfaults in C++ tests and hangs indefinitely.

PR #1513, opened 2021-03-07
HtmlRenderer: don't increment counters if src attribute is empty string

markmatney:
It may be a distant edge case for `xpp.get_attribute_value("src")` to return the empty string, but because (1) we're using it as a proxy for "attribute not found" (if that method throws `std::invalid_argument`) _and_ (2) we're already guarding against things like `<video><source><source></video>`, I figured it would be good to guard against as well.

Then, if all calls to `add_media_link` must now be wrapped in a conditional to ensure the source URL is not the empty string before incrementing the appropriate counter, the check inside the function becomes redundant.

Also adds comments to explain the use of the counters and boolean flags for media elements.

### DIFF
--- a/src/htmlrenderer.cpp
+++ b/src/htmlrenderer.cpp
@@ -276,9 +276,9 @@ void HtmlRenderer::render(std::istream& input,
 						tables.size() ? 0 : indent_level);
 
 					iframe_count++;
-
 					add_media_link(curline, links, url, iframe_url,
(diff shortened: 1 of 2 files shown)


Minoru (approved):
Thank you, very nice commens! In the future, please put the detailed explanation right into the commit message — it makes `git log` more useful because the reader doesn't have to chase GitHub PRs and issues to figure stuff out.

PR #1526, opened 2021-03-09
i18n: Add missing Italian translations for 2.23

scompo:
As for #1519 I've updated the Italian translations.


Minoru (approved):
Looks great, thank you!

PR #1538, opened 2021-03-20
Use unwrap_or instead of is_err/unwrap in utils::to_u

rnestler:
Makes the code shorter and avoids the smelly bare unwrap.


dennisschagt (approved):
Looks good, thanks!

PR #1577, opened 2021-04-05
Add support for non html mimetypes

x4e:
This PR addresses #468, #1010, and https://github.com/newsboat/newsboat/pull/1224#discussion_r502544137.

If the mime of an item's description is not one of `html`, `xhtml`, `text/html`, or `application/xhtml+xml` (given by https://tools.ietf.org/html/rfc4287#section-4.1.3.1) it will not be rendered as html, with the effect that whitespace will be properly rendered.

A follow up to this would be to add special support to render some mime types such as `text/markdown`.


x4e:
Actually I just realised there is already a PR for similar goals - #1500.
I will comment there.

Issue #1582, opened 2021-04-10
.o files of deleted .cpp files are still linked into the test executable


dennisschagt:
**Newsboat version**: 77cd6b96609688ba8910799623f50c70053e9baa (current master)

Steps to reproduce the issue:

1. Checkout a version of newsboat which does not include https://github.com/newsboat/newsboat/pull/1576, e.g. `git checkout 33716db3`

2. Run `make check` (this creates `src/logger.o`)

3. Checkout a version of newsboat which does include https://github.com/newsboat/newsboat/pull/1576, e.g. `git checkout 5cd7c179`

4. Run `make check`

Resulting errors:
```
[...]

PR #1583, opened 2021-04-10
test/test: Only link against object files of existing .cpp files

dennisschagt:
Fixes https://github.com/newsboat/newsboat/issues/1582


Minoru (approved):
Excellent idea, thanks :+1:

PR #1585, opened 2021-04-10
Prepare TagSoupPullParser for rust port 

dennisschagt:


Minoru (approved):
Thanks!

If you plan to port `TagSoupPullParser` soon-ish, please hold off until #1344 is fixed. Adding more code that passes strings over FFI will increase the visibility of that issue and make more people unhappy.

PR #1639, opened 2021-05-13
Consider toggle-article-read arguments in item view

ysh16:
as discussed in #1637
Hopefully me doing this will not create more work than necessary, the change is rather straightforward however.


Minoru (approved):
Looks great, thank you @ysh16!

PR #1642, opened 2021-05-18
Don't redraw screen when returning from temporary EmptyFormAction

dennisschagt:
Fixes https://github.com/newsboat/newsboat/issues/1463 by skipping a redraw when returning from an `EmptyFormAction`.
Before https://github.com/newsboat/newsboat/pull/1259, those lines weren't hit when opening a browser because the local `f` variable would be a `nullptr`.

### DIFF
--- a/src/view.cpp
+++ b/src/view.cpp
@@ -970,11 +970,15 @@ void View::pop_current_formaction()
 			}
 			i++;
 		}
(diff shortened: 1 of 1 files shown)


dennisschagt:
Looks like I based my changes on an old version of the master branch. I just rebased.

Minoru (approved):
Wow, that's clever! Thank you for figuring this out :)

PR #1658, opened 2021-06-06
Update Dutch translation

dennisschagt:


po/nl.po
```diff
@@ -1635,9 +1627,9 @@ msgid "`%s' is not a valid key command"
 msgstr "‘%s’ is een ongeldige toetsopdracht"
 
 #: src/keymap.cpp:775
-#, fuzzy, c-format
+#, c-format
 msgid "failed to parse operation sequence for %s"
-msgstr "Mislukt om wachtrij-bestand te openen (%s)"
+msgstr "mislukt om wachtrij-bestand te openen (%s)"
```
Minoru, line 1632: "openen" and parentheses make me think that this wasn't fully updated. Can you double-check please, @dennisschagt?

Minoru (changes requested):

po/nl.po
dennisschagt: You're right, thanks!

Fixed and force-pushed

dennisschagt:
Sorry for the two consecutive force-pushes.
I somehow overlooked the parenthesis before the first force-push.

In case it is useful:
image

Minoru:
Oh, it's fine, the patch is small enough that I could re-review the entire thing if I were paranoid :) Thanks for the quick update!

Minoru (approved):

PR #1662, opened 2021-06-08
Update German translations

der-lyse:
I not only added the missing translations, but also unified the remaining German typographic quotes (`„` and `“`) everywhere. I think I now caught all of them and there is no other quoting variant left. I searched for `` ` ``, `'`, `´` and `"`, but only the first two were present.


Minoru (approved):
Thanks! :+1:

Issue #1665, opened 2021-06-11
Newsboat crashes when entering query-feed with unknown attribute


dennisschagt:
**Newsboat version**: 8660096598c23a797a405c769565f516a09dd22a (current master)

Config file:
```
# empty
```

URLs file:
```
https://newsboat.org/news.atom newsboat
"query:invalid query:tag = \"newsboat\""
```

Steps to reproduce the issue:

1. Start newsboat

2. Select second feed and press ENTER

**Other info you think is relevant**:
![image](https://user-images.githubusercontent.com/4629607/121726951-8a384180-caeb-11eb-9428-0f7b68c60db5.png)

PR #1666, opened 2021-06-12
Reset STFL in top-level exception handlers to print in normal mode

dennisschagt:
Related to https://github.com/newsboat/newsboat/issues/1665.

Changes output on unhandled exception to:
image

Original output:
image


Minoru (approved):
Thanks!

PR #1676, opened 2021-06-19
Use cxx for FmtStrFormatter C++/rust bindings

dennisschagt:
Related issue: https://github.com/newsboat/newsboat/issues/1383

If preferred, feel free to postpone merging until after the r2.24 release.

### DIFF
--- a/rust/libnewsboat-ffi/src/fmtstrformatter.rs
+++ b/rust/libnewsboat-ffi/src/fmtstrformatter.rs
@@ -1,74 +1,38 @@
-use crate::abort_on_panic;
-use libc::{c_char, c_void};
-use libnewsboat::fmtstrformatter::FmtStrFormatter;
-use std::ffi::{CStr, CString};
-use std::mem;
+use cxx::CxxString;
+use libnewsboat::fmtstrformatter;
 
-#[no_mangle]
-pub extern "C" fn rs_fmtstrformatter_new() -> *mut c_void {
-    abort_on_panic(|| Box::into_raw(Box::new(FmtStrFormatter::new())) as *mut c_void)
+// cxx doesn't allow to share types from other crates, so we have to wrap it
+// cf. https://github.com/dtolnay/cxx/issues/496
+struct FmtStrFormatter(fmtstrformatter::FmtStrFormatter);
+
+#[cxx::bridge(namespace = "newsboat::fmtstrformatter::bridged")]
+mod bridged {
+    extern "Rust" {
+        type FmtStrFormatter;
+
+        fn create() -> Box<FmtStrFormatter>;
+
+        fn register_fmt(fmt: &mut FmtStrFormatter, key: &CxxString, value: &CxxString);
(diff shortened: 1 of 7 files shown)


src/fmtstrformatter.cpp
```diff
@@ -7,43 +7,24 @@
...
-
-FmtStrFormatter::~FmtStrFormatter()
-{
-	rs_fmtstrformatter_free(rs_fmt);
 }
 
 void FmtStrFormatter::register_fmt(char f, const std::string& value)
 {
-	rs_fmtstrformatter_register_fmt(rs_fmt, f, value.c_str());
+	std::string key(1, f);
```
Minoru, line 19: Can't the key be passed over as `u8`? If we use a one-element string, we move the knowledge from compile-time to runtime, and then have to resort to tricks like `key.chars().next()`.

rust/libnewsboat-ffi/src/fmtstrformatter.rs
```diff
@@ -1,74 +1,38 @@
...
-            CStr::from_ptr(value)
-        }
-        .to_string_lossy()
-        .into_owned();
-        // Keys are ASCII, so it's safe to cast c_char (i8) to u8 (0-127 map to the same bits).
-        // From there, it's safe to cast to Rust's char.
-        let key = key as u8 as char;
-        fmt.register_fmt(key as char, value);
+// key should contain exactly 1 char
+fn register_fmt(fmt: &mut FmtStrFormatter, key: &CxxString, value: &CxxString) {
```
Minoru, line 25: Having second thoughts about `CxxString` now. We're using it in `History` bindings already, but the rest of the codebase uses `&str`. On one hand, `CxxString` won't panic on non-UTF-8 input, but on the other hand, we now have inconsistent behaviour, which is bad.

I guess it's better to use `&str` throughout, because:

1) panics will get fixed, and then this code will have to be updated because `to_string_lossy()` doesn't exist for `&str`;
2) it's just better to have consistent behaviour, even if that behaviour is "panic on non-UTF-8 input".

Minoru (changes requested):
Thanks! I left some comments, and yeah, let's hold it until after the release — I'm not comfortable merging anything non-critical right before it.

rust/libnewsboat-ffi/src/fmtstrformatter.rs
dennisschagt: Makes sense, done (and also removed `CxxString`-usage from the `History` bindings):

src/fmtstrformatter.cpp
dennisschagt: That is definitely possible, and more in line with the original, non-CXX, FFI binding. Done: https://github.com/newsboat/newsboat/pull/1676/commits/bc6693b96a6312ab8f852a61c9f3b28429c40aef

dennisschagt:
Rebased to fix merge-conflict in mk/mk.deps:
image

Minoru (approved):
Thank you!

PR #1678, opened 2021-06-19
Fix iconv linking on OpenBSD.

botovq:
OpenBSD's libc doesn't have iconv support. Link with GNU libiconv
instead as is done for FreeBSD.


Minoru (approved):
Thanks!

If you know of any way for us to run OpenBSD in CI, please share! I looked about a year ago, and I haven't found any providers which support it.

PR #1747, opened 2021-08-22
Fix write outside of bounds in TagSoupPullParser::decode_entity

m-cz:
Adding terminating '\0' to the mbc results in a crash when pos == MB_LEN_MAX,
which is true for 4-byte characters and musl.

This is an attempt to fix it.


Minoru (approved):
Makes sense, thanks for the PR!

PR #1870, opened 2021-11-15
Add a gruvbox theme

gregf:
A simple 256-color gruvbox theme.


gregf:
img-2021-11-14-223707
img-2021-11-14-223614

Issue #1974, opened 2022-02-21
Newsboat crashes when trying to open specific feed item


mbwgh:
**Newsboat version (copy the output of `newsboat -v` or the first line of `git show`)**:

```
Newsboat 2.26.0 - https://newsboat.org/
Copyright (C) 2006-2015 Andreas Krennmair
Copyright (C) 2015-2021 Alexander Batischev
Copyright (C) 2006-2017 Newsbeuter contributors
Copyright (C) 2017-2021 Newsboat contributors

Newsboat is free software licensed under the MIT License. (Type `newsboat -vv' to see the full text.)
It bundles:
- JSON for Modern C++ library, licensed under the MIT License: https://github.com/nlohmann/json
- optional-lite library, licensed under the Boost Software License: https://github.com/martinmoene/optional-lite
- expected-lite library, licensed under the Boost Software License: https://github.com/martinmoene/expected-lite

Newsboat 2.26.0
System: Linux 5.16.10-arch1-1 (x86_64)
Compiler: g++ 11.1.0
ncurses: ncurses 6.3.20211021 (compiled with 6.3)
libcurl: libcurl/7.81.0 OpenSSL/1.1.1m zlib/1.2.11 brotli/1.0.9 zstd/1.5.2 libidn2/2.3.2 libpsl/0.21.1 (+libidn2/2.3.0) libssh2/1.10.0 nghttp2/1.46.0 (compiled with 7.80.0)
SQLite: 3.37.2 (compiled with 3.37.0)
libxml2: compiled with 2.9.12
```

Config file (copy from ~/.newsboat/config or ~/.config/newsboat/config):
Tested with an empty config file

Steps to reproduce the issue:

1. ~~Use urls file containing the single entry `https://css-tricks.com/rss`~~ Use the provided xml file and create an urls file containing the single entry `file:///path/to/css-tricks.xml` 
[css-tricks.zip](https://github.com/newsboat/newsboat/files/8114830/css-tricks.zip)

2. Create empty config / cache.db files

3. Open newsboat via `newsboat -c cache.db -C config -u urls`

4. Press R to download feed items

5. Open feed entry `A Complete Guide to CSS Cascade Layers` (if someone could tell me how to make this not dependent on remote data, I can modify this)

The following assertion then triggers:

```
[...]

PR #1975, opened 2022-02-21
Guard against popping item from empty vector (list_elements_stack)

dennisschagt:
Fixes https://github.com/newsboat/newsboat/issues/1974

### DIFF
--- a/src/htmlrenderer.cpp
+++ b/src/htmlrenderer.cpp
@@ -683,7 +683,9 @@ void HtmlRenderer::render(std::istream& input,
 						tables.size() ? 0
 						: indent_level);
 				}
-				list_elements_stack.pop_back();
+				if (!list_elements_stack.empty()) {
+					list_elements_stack.pop_back();
+				}
 				add_nonempty_line(curline, tables, lines);
 				add_line("", tables, lines);
 				prepare_new_line(curline,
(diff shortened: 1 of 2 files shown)


src/htmlrenderer.cpp
```diff
@@ -683,7 +683,9 @@ void HtmlRenderer::render(std::istream& input,
 						tables.size() ? 0
 						: indent_level);
 				}
-				list_elements_stack.pop_back();
+				if (!list_elements_stack.empty()) {
+					list_elements_stack.pop_back();
+				}
```
dennisschagt, line 688: It looks like there is some kind of rendering issue in Github's diff viewer. The indentation looks quite different from the indentation I'm seeing locally in vim (specifically the indentation of the changed code compared with the switch case indentation)
It looks like only tabs are used for indentation (as expected), so I'm not sure what is going on

image
image

Minoru: Yeah, looks like a GitHub glitch. The file is fine, so this doesn't matter :) Thanks for the heads-up though!

Minoru:
Thanks for the fix! So the issue was caused by closing a `li` tag twice? Could you add a test for that please? :) (It's fine if the test just crashes; Catch2 doesn't let us do any better).

dennisschagt:
> Thanks for the fix! So the issue was caused by closing a `li` tag twice?

It looks like it was caused by a double closed `<ul>` or `<ol>` tag.
That is also how I got this issue to reproduce in a regression test.

> Could you add a test for that please? :) (It's fine if the test just crashes; Catch2 doesn't let us do any better).

Sure, I got a bit distracted, but the requested test is now available.

Minoru (approved):
Thanks! CI failure on macOS Monterey is fixed by #1984, so I'm merging this.

PR #1995, opened 2022-03-10
Correction of the last fuzzy strings

Tonus1:
Corrected last fuzzy strings. Hope this one is ok in content and form (PR on the right branch).


der-lyse (approved):
Merci beaucoup!

PR #2076, opened 2022-05-26
Add option to center articles in ItemViewFormAction. Apply text-width when using external renderer

dennisschagt:
Resolves https://github.com/newsboat/newsboat/issues/1320

Tested using the following config with various combinations of enabled/commented-out options:
```
text-width 80
center-text yes
color end-of-text-marker black black invis
#html-renderer cat
#pager less
```

Example view:
image


doc/configcommands.dsv
```diff
@@ -141,7 +142,7 @@ ssl-verifyhost||[yes/no]||yes||If set to `no`, skip verification of the certific
 ssl-verifypeer||[yes/no]||yes||If set to `no`, skip verification of the peer's SSL certificate.||ssl-verifypeer no
 suppress-first-reload||[yes/no]||no||If set to `yes`, then the first automatic reload will be suppressed if <<auto-reload,`auto-reload`>> is set to `yes`.||suppress-first-reload yes
 swap-title-and-hints||[yes/no]||no||If set to `yes`, then the title (which is usually at the top of the screen) and the keymap hints (usually at the bottom) will exchange places. These bars can be hidden entirely, via the <<show-keymap-hints,`show-keymap-hints`>> and <<show-title-bar,`show-title-bar`>> settings.||swap-title-and-hints yes
-text-width||<number>||0||If set to a number greater than 0, all HTML will be rendered to this maximum line length or the terminal width (whichever is smaller). If set to 0, the terminal width will always be used in the article view, while <<pipe-to,`pipe-to`>>, <<save,`save`>>, and <<save-all,`save-all`>> will wrap at 80 columns instead. Does not apply when using external renderer or viewing the source. Also note that "Link" header and "Links" section won't be affected by it—they contain URLs which are better not wrapped.||text-width 72
```
Minoru, line 144: Why was this portion of the doc taken out? I think it still applies.

dennisschagt: Because it no longer applies when rendering to the screen (it did apply before I added the margins in in the STLF layout)

However, I guess it still applies when executing `save`/`save-all`/`pipe-to`, so I will re-add this text again.

Minoru: I think wrapping links in the "Links" section is a bad idea, because it breaks terminal emulators' ability to make them clickable. Same goes for the "Link" field in the header.

Wrapping in source view is less of a problem, but I still think it should be avoided too. It looks good when source HTML uses one line per paragraph, but if the source is wrapped and indented, wrapping it the second time makes it look way worse: 
newsboat

Note how every other line is extremely short — that's because the source is wrapped to something line 80 columns while I set `text-width` to 72. I realize that with `center-text yes`, switching between wrapped content and non-wrapped source would make the header jump (as the left margin disappears), but I think it's still better than the mess on the screenshot above. The feed is https://feeds.feedburner.com/tom-preston-werner if you want to test on the same input.

As for wrapping with external renderers, I'd prefer to keep the old behaviour (no wrapping) unless there is a reason to change it. That way I don't have to explain later why someone's workflow is broken :)

dennisschagt: It took me some time to see, but these are good points.

> I think wrapping links in the "Links" section is a bad idea, because it breaks terminal emulators' ability to make them clickable. Same goes for the "Link" field in the header.

I thought this would be similar to the case where the links are automatically wrapped due to a small terminal size.
However, I guess in that case the terminal allows a wrapped selection (which is not possible if it is wrapped in the middle of the terminal).
image

> Wrapping in source view is less of a problem, but I still think it should be avoided too. It looks good when source HTML uses one line per paragraph, but if the source is wrapped and indented, wrapping it the second time makes it look way worse:
&lt;image&gt;

As a side note, this also happens when the terminal is very small (even without this PR's changes), e.g.:
image

dennisschagt: I don't think these issues can be resolved while using the current approach with the STFL blocks as margins, so I will just close this PR.

@Minoru: Thank you for the review anyway!

Mati20187:
@dennisschagt was there some kind of a problem with your pull request? Because I see you've removed the branch. This is the only feature I'm missing in newsboat, that's why I'm asking

PR #2110, opened 2022-06-27
Update fr.po

Tonus1:
Getting rid of the 3 last fuzzy strings

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -267,13 +267,12 @@ msgid "%N %V - Article '%T' (%u unread, %t total)"
 msgstr "%N %V - Article '%T' (%u à lire, %t au total) - %U"
 
 #: src/configcontainer.cpp:315
-#, fuzzy
 msgid ""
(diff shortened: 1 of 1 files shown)


der-lyse (approved):
Consider my French as pretty much non-existent. However, from a punctuation point of view it's looking good to me.

Is there a subtle difference between "non lus" and "à lire" that should be kept or would it make sense to consistently go for one of them in all cases?

Minoru (approved):
Thanks! This looks good to me too.

@der-lyse, please merge once you've discussed the wording. (I'm not merging it myself now because you might decide to change it somehow.)

Tonus1:
You're right, let's keep "non lus".

I'm still not very good at choosing between beeing conservative or not for existing translation, sorry.

der-lyse:
Alright, do you want to do that in this MR or change it in a separate one, @Tonus1?

No worries, we all know translation is hard. :-)

Tonus1:
I'm quite noob with github. Already missed last release... I would be glad that these changes would me merged as soon as it could be.

Issue #2028, opened 2022-04-10
Switch to AsciiDoc keyboard macros


der-lyse:
[As it turns out](https://github.com/newsboat/newsboat/pull/2026#discussion_r846741461) there is an [AsciiDoc keyboard macro](https://docs.asciidoctor.org/asciidoc/latest/macros/keyboard-macro/), which we could use in all of our documentation to make key presses even more obvious as such.

When implementing this, make sure the man pages are also readable. Maybe some custom styling needs to be added to [our custom man page converter](https://github.com/newsboat/newsboat/blob/master/doc/man.rb), I didn't try out the `kbd:[…]` macro yet.

PR #2161, opened 2022-08-17
Use Asciidoctor's `kbd` macro to style keypresses in documentation

dennisschagt:
Resolves https://github.com/newsboat/newsboat/issues/2028

I had to add the `experimental` attribute to use the `kbd` macro.
From [the Asciidoctor documentation](https://docs.asciidoctor.org/asciidoc/latest/macros/keyboard-macro/):
image

### DIFF
--- a/doc/chapter-firststeps.asciidoc
+++ b/doc/chapter-firststeps.asciidoc
@@ -1,3 +1,5 @@
+:experimental:
+
 After you've installed Newsboat, you can run it for the first time by typing
 `newsboat` on your command prompt. This will bring you the following message:
 
@@ -74,10 +76,10 @@ The main UI of Newsboat consists of three views
 
 _Feed List View -> Article List View -> Article View_
 
-You can drill down those views by pressing `Enter` and move to the previous one by pressing
-`q`. Pressing `q` on the _Feed List View_ — or pressing `Q` from anywhere — closes Newsboat.
+You can drill down those views by pressing kbd:[Enter] and move to the previous one by pressing
+kbd:[Q]. Pressing kbd:[Q] on the _Feed List View_ — or pressing kbd:[Shift+Q] from anywhere — closes Newsboat.
 
-You can also search articles' title or content by pressing `/` on the _Feed List View_ or the _Article List View_.
+You can also search articles' title or content by pressing kbd:[/] on the _Feed List View_ or the _Article List View_.
 On the _Feed List View_ all articles of all feeds are taken into account.
 On the _Article List View_ the articles of the current feed are taken into account.
 When opening an article from a search result dialog, the search phrase is highlighted.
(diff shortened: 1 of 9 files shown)


doc/chapter-firststeps.asciidoc
```diff
@@ -143,15 +145,15 @@ Each link in the article has a number next to it.
 You can now:
 
 * Press any number to open an article link in the browser.
-  For numbers larger than 9 type `#`, then the number and press `Enter`.
-* Press `o` to open the article in the browser.
-* Press `q` to go back to the _Article List View_.
+  For numbers larger than 9 type kbd:[#], then the number and press kdb:[Enter].
```
der-lyse, line 148:
```suggestion
  For numbers larger than 9 type kbd:[#], then the number and press kbd:[Enter].
```

doc/createAvailableOperationsListView.awk
```diff
@@ -1,14 +1,38 @@
+func print_key(key) {
```
der-lyse, line 1: Unfortunately, my awk knowledge is basically non-existent. Just like CI, I also get a bunch of `return outside function body` errors. If I change `func` to `function`, it works. No idea if this is some GNU/BSD/whatever inconsistency, like different command line arguments in `grep` and friends. What ever that is worth, the GitHub highlighter also doesn't highlight the `func` keyword. My Vim, however, does (both).

    $ awk -W version
    mawk 1.3.4 20200120

der-lyse (commented):
Mate, this is already looking really awesome, I love it! :-)

Even though not finished, I thought I let you know two things. ;-)

doc/chapter-firststeps.asciidoc
dennisschagt: Thanks, fixed (amended and force-pushed)

doc/createAvailableOperationsListView.awk
dennisschagt: Thanks

For now, I got rid of the script completely (in https://github.com/newsboat/newsboat/pull/2161/commits/725699a5884ad30389808b7e5efea9013ce9eb0b) by placing the `kbd:[…]` straight into the input .dsv file (given that both of the generated files it is used for are asciidoc files)

dennisschagt:
This PR causes a slight change in style in the manpages (previously keys had underline. Now, with kbd, the keypresses look like regular text).
At first I wanted to restore the old style, but looking at it now, I actually prefer the reduced underlining in our manpages.

@der-lyse  @Minoru: Does either of you have strong feelings about this change in style?

der-lyse:
Beautiful! In the list that looks alright to. In the running text, however, I haven't decided yet if I prefer the old style with quotes or the new one without.

Old:
old

New:
new

I tried to find out how other people are doing that in their manpages, but I haven't come across anything yet. Need to do more digging. In case somebody has a bunch of reference manpages, I'm glad to hear about them. :-)

Minoru:
@der-lyse My go-tos are mutt, vim, and tmux, The first two don't mention keys at all. The latter uses single quotes and Emacs style (`'C-a x'`). Tmux approach makes sense for "chords", because they need a space to separate individual keypresses. We don't have anything like that, so I think we don't need quotes either.

@dennisschagt, @der-lyse The new style has all keys in uppercase, which looks fine without underlines/quotes to me. However, I find it a bit confusing that we use "Ctrl+X" style in docs and "^X" style in the config. I'm used to the latter, so when I see something like "Press R to reload the current feed", I press <kbd>Shift</kbd> and <kbd>r</kbd>, not just <kbd>r</kbd>.

I checked, and Asciidoc is happy with constructs like `kbd:[^C]`. They don't look as nice in HTML though, because nobody has a key with "^C" printed on it. Also, the difference between <kbd>r</kbd> and <kbd>R</kbd> might be lost on a newcomer, because they won't know that the case matters. So simply changing quotes/underlines to `kbd` won't work either.

I'm inclined to just merge this and see if users get confused too much. @der-lyse, have you made up your mind about the style?

der-lyse (approved):

Minoru:
Let's merge it then ^_^ Thanks for the work @dennisschagt!

PR #2162, opened 2022-08-20
Miscellaneous improvements

dennisschagt:
My editor was complaining about the include cycle between `view.h` and `feedlistformaction.h` (not new, it was doing that for some time already).
The include guards look okay and the compiler does not complain, so I guess it's just an over-active analyzer.
Regardless, the FeedListFormAction class was forward declared already, so I removed the superfluous include.

Afterwards, I thought it would be fun trying to minimize `view.h`'s includes.
Not sure if it's all that useful, but was a fun little exercise.


Minoru (approved):
Thanks! I read that forward declarations can speed up compilation, so I ran `include-what-you-use` a while back, but benchmarking showed no improvement. Maybe it only affects recompilations, because there are fewer accidental dependencies. I don't see harm in forward-declaring some classes though, so am happy to merge this.

PR #2165, opened 2022-08-24
Split ListWidget responsibilities for easier testing

dennisschagt:
My goals with these changes:
- Simplify testing of movement calculations for lists
- Reduce the amount of work needed to introduce new movement commands like https://github.com/newsboat/newsboat/issues/215 and https://github.com/newsboat/newsboat/issues/36

TODO:
- [x] Add tests for each of the movement options:
    - [x] on_lines_changed
    - [x] move_up
    - [x] move_down
    - [x] move_to_first
    - [x] move_to_last
    - [x] move_page_up
    - [x] move_page_down
    - [x] get_position
    - [x] set_position
- [x] ~Check if duplication in movement operation handling can now be reduced~ -> Expected to remove a bit more, however:
    - `PbView` does not inherit from `FormAction`. Can be worked around by passing "cfg" and marking the handling function static (or implementing https://github.com/newsboat/newsboat/issues/1413)
    - `DirBrowser`/`FileBrowser` have some special handling for the movement operations (e.g. changing focus to textedit if `OP_SK_DOWN` comes in while the last item of the list is selected)
    - `FeedList`/`ItemList`/`SearchResultsList` already had their movement handling in common in the `ListFormAction`, so no deduplication needed there
- [x] Verify behaviour around terminal resizes
    - => Behaviour looks the same as before refactoring

An overview, just in case it is useful:
image


dennisschagt:
I plan to write tests for the other movement methods, but I would like to first check if there are any problems with the current approach (to avoid potentially rewriting a big number of tests)

src/listwidget.cpp
```diff
@@ -2,233 +2,35 @@
...
-
-	if (get_position() >= num_lines) {
-		if (num_lines > 0) {
-			set_position(num_lines - 1);
-		} else {
-			set_position(0);
-		}
-	}
+	backend.stfl_replace_list(number_of_lines, stfl);
+	movement_control.set_position(0);
```
Minoru, line 17: Shouldn't this call `movement_control.on_lines_changed()`?

include/listwidgetbackend.h
```diff
@@ -0,0 +1,32 @@
...
+public:
+	ListWidgetBackend(const std::string& list_name, Stfl::Form& form);
+	void stfl_replace_list(std::uint32_t number_of_lines, std::string stfl);
+	void stfl_replace_lines(const ListFormatter& listfmt);
+
+	std::uint32_t get_width();
+	std::uint32_t get_height();
+	std::uint32_t get_num_lines();
+
+	void update_position(std::uint32_t pos, std::uint32_t scroll_offset);
```
Minoru, line 22: Please mark these three declarations with `override` (they're inherited from `IList`).

src/dialogsformaction.cpp
```diff
@@ -87,12 +87,12 @@ bool DialogsFormAction::process_operation(Operation op,
 {
 	switch (op) {
 	case OP_OPEN: {
-		const unsigned int pos = dialogs_list.get_position();
+		const unsigned int pos = dialogs_list.get_movement_control().get_position();
```
Minoru, line 90: I think it's ugly that the users of `ListWidget` have to know about `ListMovementControl`. How about inheriting `ListWidget` from `ListMovementControl`, so these methods become directly accessible? Then `ListMovementControl`'s interface can be split into public and protected parts, which would better explain how the class is supposed to be used. This shouldn't affect testability as it's still easy to pass a stub of `IList` in tests. What do you think?

src/dialogsformaction.cpp
```diff
@@ -103,29 +103,16 @@ bool DialogsFormAction::process_operation(Operation op,
...
-		break;
-	case OP_SK_PGDOWN:
-		dialogs_list.move_page_down(cfg->get_configvalue_as_bool("wrap-scroll"));
+		dialogs_list.get_movement_control().move_down(cfg->get_configvalue_as_bool("wrap-scroll"));
 		break;
 	case OP_QUIT:
 		v->pop_current_formaction();
 		break;
 	default:
+		handle_list_operations(dialogs_list.get_movement_control(), op);
```
Minoru, line 115:
```suggestion
		if (handle_list_operations(dialogs_list.get_movement_control(), op)) {
			break;
		}
```

(ditto in other places that call `handle_list_operations`)

This will establish a pattern of extending the `default` clause with more such methods, and ensure that only one method handles a particular operation.

Minoru (commented):
I like the idea, but have a couple comments on technicalities :)

dennisschagt:
Good ideas, I just force-pushed to make sure this PR can get merged properly (given that the dependency list needed updating, but is already removed from the master branch)

src/dialogsformaction.cpp
dennisschagt: Thanks, I like it. It makes it immediately obvious how to extend this if there is any more de-duplication effort

src/dialogsformaction.cpp
dennisschagt: Sounds good.
However, I just tried to execute this idea and ran into some issues.

The `ListMovementControl` requires the `IList` reference directly in its constructor (if we want to store it as a reference in a field).
That gives some issues in the `ListWidget` if it derives from the `ListMovementControl`.
Deriving it requires that the `ListMovementControl` is initialized first, but it already needs access to the `IList` (`ListWidgetBackend`) which is a field in the later constructed `ListWidget`.
Other related options I thought about had similar issues (or created a lot of ugliness in the designs of the classes)

I did find a way to get rid of all the `get_movement_control()` calls by providing dereferencing operators (`operator->` and `operator*`.
I think it looks quite natural in the places where it is used.
Do you find this solution acceptable?

include/listwidgetbackend.h
dennisschagt: Good catch, done

src/listwidget.cpp
dennisschagt: Oops, just missed this while applying+pushing the requested changes.

It is not really necessary, given that `movement_control.set_position(0)` will move both the position and scroll_offset to 0 (which is the old behavior).
However, I guess it makes sense to always call `on_lines_changed` to make sure this keeps working even if more internal state is added to the movement control

src/dialogsformaction.cpp
Minoru: Oh, my bad, I should have realized that the initialization order would mess that up!

If the call to `get_movement_control()` was *too* explicit to my taste, then the call to `operator->` is too implicit :) It looks quite mysterious and unexpected that I have to dereference the object in order to get to some of its methods.

I have two more ideas though. If you aren't interested, then please roll back to `get_movement_control()` and continue with that.

The first idea is to make `ListMovementControl` a mixin: replace `IList& list` member with `protected: virtual IList& get_list() = 0`, then inherit `ListWidget` from it and implement the method as `ILIst& get_list() override { return backend; }`.

The second idea requires more changes, but actually results in less code than any of the previous ones:

1. make `ListMovementControl` a template, parameterized by the backend type. Inherit `ListMovementControl` from the backend type, both to get to its methods and "re-export" its public interface;
2. `ListWidget` becomes `using ListWidget = ListMovementControl<ListWidgetBackend>;` 
3. backend methods that are only relevant to movement control can become `protected` now;
4. drop `IList` because it has no purpose anymore.

In essence, we build up `ListWidget`'s interface from the ground up: `ListWidgetBackend` gives us the two `stfl_replace_*` methods and `get_width()`, and then `ListMovementControl` gives us the movement methods.

What do you think about those?

dennisschagt: No problem, always nice trying out new things :)
I just pushed an attempt at implementing your second idea.
Is this what you had in mind?

I might need to take another look at the current tests to see if they can be cleaned up a bit (for now, I just stopped as soon as I got them to compile again)
However, I probably won't have time until next weekend.

dennisschagt:
> BTW I don't mind rebasing at this stage, because half the changes are mechanical. And if one of my ideas pans out, these changes will go away too... ;)

I just rebased onto master:
image

Let me know if you want me to squash everything together.

include/listmovementcontrol.h
```diff
@@ -0,0 +1,207 @@
...
+	// Make sure the backend's destructor is marked virtual
+	~ListMovementControl() override = default;
+
+	void set_num_context_lines(std::uint32_t context_lines)
+	{
+		num_context_lines = context_lines;
+
+		// Force a recalculation of the scroll offset
+		set_position(current_position);
+	}
```
Minoru, line 29: This method is always called after the constructor, and is never called anywhere else. I suggest replacing it with a constructor:

```c++
ListMovementControl::ListMovementControl(const std::string& list_name, Stfl::Form& form,
	std::uint32_t scrolloff)
    : Backend(list_name, form)
{
    num_context_lines = scrollof;
    set_position(current_position);
}
```

`Backend`'s constructor can be made `protected` to avoid people calling it by accident.

src/dialogsformaction.cpp
Minoru: Yep, looks precisely as I envisioned it, great job!

Minoru:
Yeah, squashing these into a single commit looks like the right thing to do.

include/listmovementcontrol.h
dennisschagt: Makes sense, done

dennisschagt:
> Yeah, squashing these into a single commit looks like the right thing to do.

Rebased and squashed.

I guess I shouldn't have rebased.
In case it is useful, this is the diff for the new push if I hadn't rebased: https://github.com/newsboat/newsboat/compare/ce00d60f3b52488af915243b44f184605533b060..af76f4172cc2f0510dde9bdbd4a719363d0d8186

include/listwidget.h
```diff
@@ -2,46 +2,15 @@
 #define NEWSBOAT_LISTWIDGET_H_
 
 #include <cstdint>
-#include <memory>
 #include <string>
 
 #include "listformatter.h"
-#include "stflpp.h"
+#include "listmovementcontrol.h"
+#include "listwidgetbackend.h"
```
Minoru, line 9: Please remove all other `#include`s so Make doesn't rebuild stuff unnecessarily.

Minoru (approved):
Cool, thank you! I found a few minor problems in tests, please merge this after fixing them.

test/listmovementcontrol.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+			THEN("the 6th item is selected") {
+				REQUIRE(moved == true);
+				REQUIRE(list.position == 5);
+				REQUIRE(list.scroll_offset == 0);
+			}
+		}
+
+		WHEN("move_down() is called with wrapping enabled") {
+			bool moved = list_movement.move_down(false);
```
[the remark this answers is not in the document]
dennisschagt: Oops, I guess I missed a lot of these. Thanks!

PR #2185, opened 2022-09-12
Updated italian translations

scompo:
Updated the italian po file


dennisschagt (approved):
Looks good, thank you!

PR #2190, opened 2022-09-16
chore(l10n): Update pl translation before release

carno:
Ref: https://github.com/newsboat/newsboat/pull/2180#issuecomment-1243021481


der-lyse (approved):
LGTM, thank you!

PR #2218, opened 2022-10-16
Accept enclosure with invalid type if no valid enclosure is found

dennisschagt:
Resolves https://github.com/newsboat/newsboat/issues/2050

This implements @Minoru's idea from https://github.com/newsboat/newsboat/issues/2050#issuecomment-1114181986:
> So Newsboat is technically correct, but I still think it could do better :) Current algorithm is to pick the last enclosure of the supported type. I guess it could be changed to pick the last enclosure (regardless of the type) if the existing algorithm hasn't picked anything.


Minoru (approved):
Nice, thanks a lot! This even paves way for support of multiple enclosures :rocket:

PR #2269, opened 2022-11-28
Fix clone command

pepellou:
This would be the right syntax when using the `git` protocol.

Alternatively, we could have kept the existing url and changed the protocol to `https`, but I think probably `git` is the preferred (to force the use of SSH key).

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ https://newsboat.org/
 
 Alternatively, you can check out the latest version from the Git repository:
 
-	$ git clone git://github.com/newsboat/newsboat.git
+	$ git clone https://github.com/newsboat/newsboat.git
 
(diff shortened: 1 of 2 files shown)


Minoru (changes requested):
Ah, I missed [the blog post](https://github.blog/2021-09-01-improving-git-protocol-security-github/) that announced the removal of unencrypted git:// support from GitHub. Nice catch! A couple notes though:

1. the README shouldn't assume that the user has an SSH key with GitHub, so please change the command to use HTTPS;
2. please update _doc/newsboat.asciidoc_ to match.

pepellou:
Good point @Minoru , PR updated 👍

Minoru (approved):

PR #2309, opened 2023-01-08
add missing <cstdint> includes for gcc-13 compatibility

trofi:
Upcoming `gcc-13` made `<string>` leaner and does not include `<cstdint>`
implicitly anymore. As a result build fails without the change as:

    include/matcherexception.h:14:14: error: elaborated-type-specifier for a scoped enum must not use the «class» keyword [-Werror]
       14 |         enum class Type : std::uint8_t { ATTRIB_UNAVAIL = 0, INVALID_REGEX = 1 };
          |         ~~~~ ^~~~~
          |              -----


dennisschagt (approved):
Looks good, thank you!

PR #2339, opened 2023-02-12
Add support for cmd-one..cmd-nine in dialog overview

dennisschagt:


Minoru (approved):
Thanks! I think the "Command Line" section in the docs should be updated to mention that dialogs view support these keys now. Feel free to merge once that's done.

dennisschagt:
Good point, thanks for the review!

dennisschagt:
The coverage build is currently broken.
I can reproduce the issue locally by running `cargo +nightly install --locked grcov --force` (after updating to the latest Rust nightly toolchain)

The build fails on nightly with `proc-macro2` versions before v1.0.60.
grcov's Cargo.lock file specifies `proc-macro2` v1.0.52.
We now force using that by specifying the `--locked` option when running `cargo install grcov`.

@Minoru Is there a specific reason why we specify `--locked` there? (Can I remove it?)

Related issue on grcov repo: https://github.com/mozilla/grcov/issues/1069

dennisschagt:
I guess I should have looked at the `git blame` output: https://github.com/newsboat/newsboat/commit/23b7c6d0cbf38cf2bdd0b547d8c70be1bdcadf20
It looks like the `--locked` option was only added as a workaround for a different issue (which I expect to be no longer an issue).

PR #2340, opened 2023-02-12
Convert feed into utf-8 if charset is specified in HTTP Content-Type header

dennisschagt:
Previous attempts at implementing encoding fixes (https://github.com/newsboat/newsboat/pull/2214 + https://github.com/newsboat/newsboat/pull/2243) had some  issues when encodings were specified both in the HTTP headers and in the XML header.
https://github.com/newsboat/newsboat/commit/fa363edd2d61e598aff0de5848ae955a5b35d3c0 reverts those problematic changes.
In this PR, I redo most of the reverted changes while implementing the idea proposed near the end of https://github.com/newsboat/newsboat/issues/1436#issuecomment-1367995038.

For now, I've only checked the `Content-Type` header in `rsspp::Parser::parse_url()`.
I prefer to fix the other places (at least `Utils::retrieve_url()`) in a separate PR.

There are a few more locations where HTTP transfers occurs, mainly in the external APIs.
I prefer to also pick those up in a separate PR (maybe letting them reuse the `Utils::retrieve_url()`, tracked in https://github.com/newsboat/newsboat/issues/2227).

Related issue:
- https://github.com/newsboat/newsboat/issues/1436 (this PR fixes it, but it might be useful to keep it open until we take care of the Content-Type in all relevant locations)

### DIFF
--- a/include/curlheadercontainer.h
+++ b/include/curlheadercontainer.h
@@ -11,7 +11,7 @@ namespace newsboat {
 
 class CurlHeaderContainer {
 public:
-	[[nodiscard("Registration is only valid as long as the pointer is alive")]]
+	// Registration is cleaned up on destruction
 	static std::unique_ptr<CurlHeaderContainer> register_header_handler(
 		CurlHandle& curlHandle);
 
--- a/include/curlheadercontainer.h
+++ b/include/curlheadercontainer.h
@@ -16,6 +16,7 @@ public:
 		CurlHandle& curlHandle);
 
 	const std::vector<std::string>& get_header_lines() const;
+	std::vector<std::string> get_header_lines(const std::string& key) const;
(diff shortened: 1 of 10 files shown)


dennisschagt:
Rebased to get updated Cirrus CI configurations and to resolve a conflict with changes from https://github.com/newsboat/newsboat/pull/2395.

dennisschagt:
Now that https://github.com/newsboat/newsboat/pull/2752 is merged, I think it is possible to write some tests for the changes in this PR.
That might help in getting a bit more confidence in the correctness of this solution.

dennisschagt:
I think I will abandon this PR and instead fix the issues using the pseudo-code from https://github.com/newsboat/newsboat/pull/2660#discussion_r1449461867

Other improvements in this PR are split of already.

PR #2356, opened 2023-02-28
Update ES translation

Roboron3042:
Didn't come in time for the release this time, but better late than never!


der-lyse (approved):
Thank you!

PR #2374, opened 2023-03-15
Update German translations

der-lyse:
The German language allows short list items to start in lower or uppercase. I could not figure out if any version is slightly preferred over the other. Most of the keymap entries were using lowercase, however, some also started in uppercase (not counting for nouns which are always capitalized). As the original English descriptions also use uppercase I decided to switch to uppercase in order to help avoiding mixed styles in the future for new keymap entries.

Also, I noticed that two different forms for "previous" ("vorherig" and "vorhergehend") had been used, so I normalized them all to the shorter version.

Finally, there was a typo in "als" where the "l" was missing. I'm only mentioning this to avoid potential confusion with the review by any non-German reviewers.


dennisschagt (approved):
Nice improvements, looks good to me

PR #2375, opened 2023-03-15

carno:
Ref: https://github.com/newsboat/newsboat/pull/2371#issuecomment-1470827042

### DIFF
--- a/po/pl.po
+++ b/po/pl.po
@@ -6,7 +6,7 @@ msgstr ""
 "Project-Id-Version: newsboat 2.5\n"
 "Report-Msgid-Bugs-To: https://github.com/newsboat/newsboat/issues\n"
 "POT-Creation-Date: 2023-03-15 23:28+0300\n"
-"PO-Revision-Date: 2022-12-12 22:20+0100\n"
+"PO-Revision-Date: 2023-03-16 21:55+0100\n"
 "Last-Translator: Michal Siemek <carnophage@dobramama.pl>\n"
 "Language-Team: Polish <pl@li.org>\n"
 "Language: pl\n"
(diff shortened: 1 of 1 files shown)


po/pl.po
```diff
@@ -98,7 +98,9 @@ msgid ""
 "write a log with a certain log level (valid values: 1 to 6, for user error, "
 "critical, error, warning, info, and debug respectively)"
 msgstr ""
-"utwórz log używając wybranego poziomu logowania (poprawne wartości: 1 do 6)"
+"utwórz log używając wybranego poziomu logowania (poprawne wartości: 1 do 6, "
+"dla błędów użytkownika, krytycznych, błędów, ostrzeżeń, informacji i "
```
carno, line 102: Honestly I'm not really happy with this translations. It sounds weird to translate log levels and I have no sensible translation for `debug` :(

der-lyse: I hear you! Had the same trouble with the German translations and just went with the English name in the end. Most if not all users will know what that means.

Line 96 with the "fuzzy" should be removed, I reckon.

carno: For now I have left `fuzzy` on purpose to give myself some time to think about it. Maybe I will stick to leaving log levels untranslated too.

der-lyse (approved):
Thanks!

For what it's worth, you previously had some translations for the other log levels. Not sure if it makes sense to translate "user error", "critical error" etc. into Polish and only keep the English "debug". If I'm not mistaken, that's what the already submitted translations do. As I don't speak all those languages, I can't tell whether in those languages "debug" natively exists or is an English loanword.

But you're the Polish expert here (I have absolutely no clue), if your current version is good enough, that's perfectly fine with me. :-) Happy to merge as is.

carno:
> But you're the Polish expert here (I have absolutely no clue), if your current version is good enough, that's perfectly fine with me. :-) Happy to merge as is.

Well I definitely don't feel like one ;-)

OK, so here is my train of thought and why I would go with current approach. Most of the words describing log levels can be translated (and `to debug` is basically `debugować`), but I don't believe we often use translated words for describing log levels, it might be my bias from working in IT, but I find English names for log levels more familiar.

der-lyse:
> Well I definitely don't feel like one ;-)

Hahaha. :-D At least to me you are. :-)

> […] I don't believe we often use translated words for describing log levels, it might be my bias from working in IT, but I find English names for log levels more familiar.

Alright, thanks! Let's merge it.

Issue #2367, opened 2023-03-13
Images in Mastodon Feeds are recognized as Podcast and get auto-enqueued


dorsiflexion:
**Newsboat version**:

```
Newsboat 2.30.1
System: Linux 6.1.15-hardened1-1-hardened (x86_64)
Compiler: g++ 12.2.0
ncurses: ncurses 6.4.20221231 (compiled with 6.3)
libcurl: libcurl/7.88.1 OpenSSL/3.0.8 zlib/1.2.13 brotli/1.0.9 zstd/1.5.4 libidn2/2.3.4 libpsl/0.21.2 (+libidn2/2.3.4) libssh2/1.10.0 nghttp2/1.52.0 (compiled with 7.87.0)
SQLite: 3.41.1 (compiled with 3.40.0)
libxml2: compiled with 2.10.3
```

**Config file**:

```
urls-source "ocnews"
ocnews-url ***
ocnews-login ***
ocnews-password ***
ssl-verifypeer no
ssl-verifyhost no
podcast-auto-enqueue yes
```

**Steps to reproduce the issue**:

1. Add a Mastodon feed to Nectcloud news.
2. The article must not be in Newsboat cache. Retrieve Nextcloud news with Newsboat – i. e. open Newsboat and reload feed.
3. Auto-enqueue should already have enqueued the image to the podboat podcast queue.

**Other info you think is relevant**:

Feed URL: https://social.tchncs.de/@kuketzblog.rss
Toot URL: https://social.tchncs.de/@kuketzblog/109975086360478784

The full article as shown by newsboat:

```
Feed: Kuketz IT-Security » Mastodon
Title: 109975086360478784
Date: Mon, 06 Mar 2023 08:10:44 +0100
Link: https://social.tchncs.de/@kuketzblog/109975086360478784
Podcast Download URL:
https://f2.tchncs.de/media_attachments/files/109/975/078/545/360/548/original/5d55b6d57093f18a.png (type:
image/png)

Nachdem das neueste Update von GrapheneOS eingespielt ist, kann man unter
"Einstellungen -> Standort" nun den SUPL-Proxy-Server auswählen. 👇
```

**Comment**:

What is the logic behind auto-enqueueing? Does every attachment get enqueued, no matter what?

I think newsboat should at least check if the file is an audio file.

dennisschagt:
> What is the logic behind auto-enqueueing? Does every attachment get enqueued, no matter what?

[...]

Issue #2305, opened 2022-12-30
Data loss observed in importing data from Mastodon RSS Feeds


sgharms:
Hello, I'm seeing the following data loss for subscriptions to RSS publications of Mastodon users' feeds. My goal is to make newsboat my primary interface into Mastodon. Here's a Mastodon page:

Behold, the adoable Latin-language account for the mouse Minimus:

<img width="801" alt="image" src="https://user-images.githubusercontent.com/50019/210044671-25f0f48d-eb57-4f64-a209-0d527f0bf6b1.png">

For copy/paste ease, the URL is: https://archaeo.social/@Minimus/109597925165056089

This article shows up in Newsboat like:

<img width="1175" alt="image" src="https://user-images.githubusercontent.com/50019/210044747-ce5e8ad8-6da0-45d1-972d-0ac29a4e4728.png">

Note, there's no hint of the `<img>` payload 😢 

Viewing the source with `^U` just shows the same text content but with the wrapping `<p>` tags. Using the `show-urls` function does not see a URL that points to the image.

Checking the cache database:

```sql
sqlite> SELECT * FROM rss_item WHERE url='https://archaeo.social/@Minimus/109597925165056089';
10329|https://archaeo.social/@Minimus/109597925165056089|109597925165056089||https://archaeo.social/@Minimus/109597925165056089|https://archaeo.social/@Minimus.rss|1672331621|<p>Time to start writing those thank you letters, Minimus!</p>|0|||0||0||
```

OK, so in the cache there is clearly no mention of the image. However, if I go to:

https://archaeo.social/@Minimus.rss and extract the raw XML for this post:

```xml
    <item>
      <guid isPermaLink="true">https://archaeo.social/@Minimus/109597925165056089</guid>
      <link>https://archaeo.social/@Minimus/109597925165056089</link>
      <pubDate>Thu, 29 Dec 2022 16:33:41 +0000</pubDate>
      <description>&lt;p&gt;Time to start writing those thank you letters, Minimus!&lt;/p&gt;</description>
      <media:content url="https://mstdn.archaeo.social/media_attachments/files/109/597/921/900/998/086/original/fe1c399c59936269.jpeg" type="image/jpeg" fileSize="376093" medium="image">
        <media:rating scheme="urn:simple">nonadult</media:rating>
        <media:description type="plain">Minimus, the Latin mouse, writing on a wax tablet</media:description>
      </media:content>
    </item>
```

[...]

PR #2378, opened 2023-03-16
Only autoenqueue podcasts with a valid or empty type

dennisschagt:
Fixes https://github.com/newsboat/newsboat/issues/2367

### DIFF
--- a/src/queuemanager.cpp
+++ b/src/queuemanager.cpp
@@ -122,10 +122,19 @@ EnqueueResult QueueManager::autoenqueue(std::shared_ptr<RssFeed> feed)
 			continue;
 		}
 
+		const auto enclosure_type = item->enclosure_type();
+		const auto enclosure_url = item->enclosure_url();
+
+		if (!enclosure_type.empty() && !utils::is_valid_podcast_type(enclosure_type)) {
+			LOG(Level::DEBUG, "QueueManager::autoenqueue: Skipping enclosure with url `%s'"
+				" becausoe of invalid podcast type `%s'", enclosure_url, enclosure_type);
+			continue;
+		}
+
 		LOG(Level::DEBUG,
 			"QueueManager::autoenqueue: enclosure_url = `%s' enclosure_type = `%s'",
-			item->enclosure_url(),
-			item->enclosure_type());
+			enclosure_url,
+			enclosure_type);
 		if (utils::is_http_url(item->enclosure_url())) {
 			LOG(Level::INFO,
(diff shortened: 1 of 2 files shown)


src/queuemanager.cpp
```diff
@@ -122,10 +122,19 @@ EnqueueResult QueueManager::autoenqueue(std::shared_ptr<RssFeed> feed)
 			continue;
 		}
 
+		const auto enclosure_type = item->enclosure_type();
+		const auto enclosure_url = item->enclosure_url();
+
+		if (!enclosure_type.empty() && !utils::is_valid_podcast_type(enclosure_type)) {
+			LOG(Level::DEBUG, "QueueManager::autoenqueue: Skipping enclosure with url `%s'"
+				" becausoe of invalid podcast type `%s'", enclosure_url, enclosure_type);
```
der-lyse, line 130:
```suggestion
				" because of invalid podcast type `%s'", enclosure_url, enclosure_type);
```

dennisschagt: Oops, fixed

Minoru (changes requested):
Sorry I'm so late with the review!

I agree with the new `if` condition, but I think it should be in `RssParser`. In #2218 individual parsers (RSS 1.0, Atom etc.) learned to extract *all* the enclosures, and `RssParser` was made responsible for picking the "best" enclosure. I think we should just to modify the algorithm there.

dennisschagt:
No problem, thank you for the review anyway!

> In https://github.com/newsboat/newsboat/pull/2218 individual parsers (RSS 1.0, Atom etc.) learned to extract all the enclosures, and RssParser was made responsible for picking the "best" enclosure. I think we should just to modify the algorithm there.

I prefer to keep the decision of which items to (auto-)enqueue in the queuemanager.
My idea for enclosures and other media attachments is as follows:
- Support multiple attachments (both enclosures and `media:content` elements, removing the need to pick the "best" enclosure)
- Enqueue only items with audio/video/unknown mime-type (specifically, don't enqueue images => check added in this PR)
- Show image attachments near the top of the itemview (between the headers and the actual text, rendering them similar to links specified in item's HTML content)
- Related idea: https://github.com/newsboat/newsboat/issues/2305#issuecomment-1493412650

To be able to show images, we should not already filter them out in the `RssParser`.

Does that make sense to you / do you agree?

Minoru (approved):
Yup, makes total sense! I failed to consider how we could evolve this.

PR #2426, opened 2023-05-19
PbView: Change some pointers to references

dennisschagt:
Just a small improvement introduced in https://github.com/newsboat/newsboat/pull/2364 (splitting it off so the other PR becomes slightly easier to review)


Minoru (approved):
Thanks!

PR #2475, opened 2023-06-25
Completing and fixing the Swedish translation

deob83:
This fixes the unresolved issues from #2471, but also completes the Swedish translation as well as fixing/tweaking some previously translated strings. 

(Note: some of the diffs here comes from Poedit changing the position of line breaks in the file.)


der-lyse (approved):
Thank you! Let's hope that with the next update of the po file when extracting the strings from the source files the line breaks do not change back. :-) (I have no idea.)

PR #2479, opened 2023-06-29
Update ES translation

Roboron3042:


der-lyse (approved):
Muchas gracias. I don't speak Spanish, but this looks alright to me.

PR #2489, opened 2023-07-09
Teach contrib/image-preview/nbrun to take Newsboat's CLI arguments 

venomega:


contrib/image-preview/nbrun
```diff
@@ -11,5 +11,5 @@ mkfifo "$FIFO_UEBERZUG"
 trap cleanup EXIT
 tail --follow "$FIFO_UEBERZUG" | ueberzug layer --silent --parser bash &
 
-newsboat
+newsboat $@
```
dennisschagt, line 14: It might be useful to put the `$@` within double quotes:
```suggestion
newsboat "$@"
```

See for some discussion:
- https://stackoverflow.com/questions/3811345/how-to-pass-all-arguments-passed-to-my-bash-script-to-a-function-of-mine/3816747#3816747
- https://stackoverflow.com/questions/4824590/propagate-all-arguments-in-a-bash-shell-script/4824637#4824637

Other than that, looks like a nice/useful change.

venomega:
i know about "$@" but i tested that way and it worked, i thought that "$@"
is just for arguments with '\ ' on it

Minoru:
`"$@"` is for arguments with spaces and wildcards in them. I haven't tested nbrun myself, but in general I think it's better to escape the variable, so I'll commit @dennisschagt's suggestion and will merge this. Thank you for the pull request!

Issue #2541, opened 2023-09-07
Links summary not rendered for plaintext articles


dennisschagt:
**Newsboat version**: r2.32-136-g79119

Steps to reproduce the issue:

1. Add a youtube feed to the urls file

2. Open one of the items

3. Notice `Link[1]` in the rendered item, but no link overview at the bottom

![image](https://github.com/newsboat/newsboat/assets/4629607/ce8d549e-f108-4920-988a-da34e7499a74)

**Other info you think is relevant**:
Extra link introduced by https://github.com/newsboat/newsboat/pull/2399
I think this can be resolved by moving the rendering of the links summary from the `HtmlRenderer` to the `ItemRenderer`:
https://github.com/newsboat/newsboat/blob/79119120812e409b72823ac42979d16b46c94b41/src/htmlrenderer.cpp#L1016-L1027

PR #2543, opened 2023-09-09
Always render links if there are any

dennisschagt:
Mainly moved code from `HtmlRenderer` to `Links` and `ItemRenderer`.
`ItemRenderer` already had some tests with rendered links.

Fixes https://github.com/newsboat/newsboat/issues/2541

### DIFF
--- a/include/links.h
+++ b/include/links.h
@@ -1,9 +1,14 @@
 #ifndef NEWSBOAT_LINKS_H_
 #define NEWSBOAT_LINKS_H_
+
 #include <vector>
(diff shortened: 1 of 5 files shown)


Minoru (changes requested):
Nice work!

Can you please add a test for the original bug, i.e. an item with an image enclosure and plaintext content?

dennisschagt:
Sure, done.
Also verified that this same new test fails on the current master branch.

Minoru (approved):
Thanks!

PR #2555, opened 2023-09-16
Prepare for "bind" command by specifying binding type instead of boolean "automatic"

dennisschagt:
This implements an idea by @Minoru in https://github.com/newsboat/newsboat/pull/2507#issuecomment-1715312777

This is done in preparation of adding a new "bind" command.
We would like to handle some operations differently without changing the behaviour for the current bind-key and macro bindings.

This obsoletes https://github.com/newsboat/newsboat/pull/2507
I've cherry-picked one of the commits:
> (if we end up closing this PR) https://github.com/newsboat/newsboat/commit/c51cabb77fd62c2dfa17ee1b080ac4c6ca960849 is a very good observation and I'd like to have it anyway.

### DIFF
--- a/include/dialogsformaction.h
+++ b/include/dialogsformaction.h
@@ -29,7 +29,7 @@ protected:
 
 private:
 	bool process_operation(Operation op,
-		bool automatic = false,
+		BindingType bindingType = BindingType::BindKey,
 		std::vector<std::string>* args = nullptr) override;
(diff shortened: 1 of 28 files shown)


dennisschagt:
Good idea, done

Minoru (approved):
Thanks!

PR #2607, opened 2023-12-01
Automation updates of `metadata` table 

bogdasar1985:
Issue https://github.com/newsboat/newsboat/issues/2528

### DIFF
--- a/src/cache.cpp
+++ b/src/cache.cpp
@@ -437,10 +437,11 @@ void Cache::populate_tables()
 			std::to_string(patch_version.major) + ", db_schema_version_minor = " +
 			std::to_string(patch_version.minor) + ";";
 
+		if (patch_version > SchemaVersion{2, 11}) {
+			run_sql_nothrow(update_metadata_query);
+		}
+
 		for (const auto& query : patches_it->second) {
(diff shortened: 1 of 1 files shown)


src/cache.cpp
```diff
@@ -435,7 +432,15 @@ void Cache::populate_tables()
...
 			patch_version.major,
 			patch_version.minor);
+
+		std::string update_metadata_query = "UPDATE metadata SET db_schema_version_major = " +
+			std::to_string(patch_version.major) + ", db_schema_version_minor = " +
+			std::to_string(patch_version.minor) + ";";
+
 		for (const auto& query : patches_it->second) {
+			if (patch_version > SchemaVersion{2, 11}) {
+				run_sql_nothrow(update_metadata_query);
```
dennisschagt, line 442: I think this only needs to be run once per set of SQL queries, so just outside of this inner for-loop.

dennisschagt (commented):
Nice improvement!

src/cache.cpp
bogdasar1985: @dennisschagt that's right

dennisschagt (approved):
Thanks, LGTM

PR #2608, opened 2023-12-06
Update Chinese translations

CookiePieWw:


po/zh.po
```diff
@@ -1919,120 +1877,114 @@ msgstr "错误：无法将种子标记为已读: %s"
...
+msgstr "%s: %s: 无效的日志等级"
 
 #: rust/libnewsboat/src/configpaths.rs:85
-#, fuzzy
 msgid ""
 "Fatal error: couldn't determine home directory!\n"
 "Please set the HOME environment variable or add a valid user for UID %u!"
-msgstr "请设置主目录的环境变量，或者添加一个有效的用户其UID为 %u!"
+msgstr "严重错误：无法确定家目录！\n"
+"请设置主目录的环境变量，或者添加一个有效的用户其UID为 %u!"
```
der-lyse, line 1923: I reckon the `HOME` variable should be included here, too. Since it's actually called `HOME`, it should not be translated.

po/zh.po
```diff
@@ -1919,120 +1877,114 @@ msgstr "错误：无法将种子标记为已读: %s"
...
 #: rust/libnewsboat/src/filterparser.rs:131
 msgid "one of: =~, ==, =, !~, !=, <=, >=, <, >, between, #, !#"
-msgstr ""
+msgstr "其中之一：=~、==、=、!~、!=、<=、>=、<、>、between、#、!#"
 
 #. The options ("quoted string" etc.) are not keywords, so please translate them.
 #: rust/libnewsboat/src/filterparser.rs:133
 msgid "one of: quoted string, range, number"
-msgstr ""
+msgstr "其中之一：quoted string、range、number"
```
der-lyse, line 1949: In case there is a good way to express "quoted string", "range" and "number" in Chinese, please translate them. You _might_ have moved the "not" from the programmer's hint from the first into the second half of the sentence in your head. :-)

po/zh.po
```diff
@@ -1919,120 +1877,114 @@ msgstr "错误：无法将种子标记为已读: %s"
...
 #: rust/libnewsboat/src/filterparser.rs:437
 msgid "Internal parse error"
-msgstr ""
+msgstr "内部解析错误"
 
 #: rust/libnewsboat/src/fslock.rs:57
-#, fuzzy
 msgid "Failed to open lock file '%s': %s"
-msgstr "当调用`%s'的时候出错: %s"
+msgstr "打开锁文件 '%s' 失败: %s"
```
der-lyse, line 1969: In the translations above, ``` `%s' ``` was not surrounded by spaces. Now with `'%s'` there are spaces placed around it. This looks inconsistent to me (and yes, Newsboat is inconsistent with the used types of quotes, too). We can postpone the decision and you make up your mind which way would be best for Chinese and then fix it in a future merge request. Up to you. :-)

po/zh.po
```diff
@@ -1324,29 +1303,24 @@ msgid "Go to a random unread article"
 msgstr "转到随机一篇未读文章"
 
 #: src/keymap.cpp:126
-#, fuzzy
 msgid "Open URL of article, or entry in URL view. Mark read"
-msgstr "在浏览器里打开文章"
+msgstr "在链接视图(URL view)中打开文章或某一条目(entry)。标记为已读"
```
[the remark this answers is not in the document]
CookiePieWw: They are both translated and I placed the original English texts after their translations with `()`.

`链接视图` for `URL view` and `条目` for `entry` are not frequently used in Chinese, so it may be confusing when first encountered. From my point of view, I would like to check over the docs to find the original usage of these (we can say terminologies?). Since Newsboat hasn't had docs in other languages, I think it's better to help the user find the terms in the English docs instead of getting stuck in the Chinese-English matching.

The same question goes to `query feeds` which is more like a term. There is a [chapter](https://newsboat.org/releases/2.33/docs/newsboat.html#_query_feeds) in the docs to explain it. If I translate it like this: `查询` for `query` and `种子` for `feeds`,  so `查询种子`, it may be more confusing than `链接视图` and `条目` above. I haven't decided how to translate it yet.

po/zh.po
```diff
@@ -1614,72 +1579,66 @@ msgid "Start cmdline with 9"
...
 msgid "`%s' is not a valid key command"
-msgstr ""
+msgstr "`%s' 不是一个有效的键盘指令"
 
 #: src/keymap.cpp:764
-#, fuzzy, c-format
+#, c-format
 msgid "failed to parse operation sequence for %s"
-msgstr "当调用`%s'的时候出错: %s"
+msgstr "解析操作序列失败：%s"
```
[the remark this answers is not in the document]
CookiePieWw: > it could be just a relic from the previous fuzzy version

Yes, it is a relic and also an acceptable translation. I think `%s` refers to the text of the `operation sequence`, so it's also alright to place it just after the `操作序列` which means `operation sequence`. Updated in the latest commit.

po/zh.po
CookiePieWw: You're right. I missed all `HOME` at first, fixed some later but missed this one. Updated in the latest commit.

po/zh.po
CookiePieWw: You're right here. Updated in the latest commit.

po/zh.po
CookiePieWw: It's a habit to have spaces between English letters and Chinese words for readability. I focused on the parts with `fuzzy` tags this time so I failed to concentrate on styles. Besides, I noticed the previous translations that I didn't change are also inconsistent, so maybe it's better to create a PR focusing on styles later?

der-lyse (approved):
Awesome, thanks for all your work! Looking forward to future improvements. :-) But I know too well from my efforts on the German version that it takes a long time to come up with good translations and making them consistent in style and terminology. So, take your time. It's an endless endeavor. ;-)

PR #2617, opened 2023-12-11
 chore(l10n): Update Polish translation before 2.34 release

carno:
Ref: https://github.com/newsboat/newsboat/pull/2616#issuecomment-1850666683


der-lyse (approved):
Thanks, looks good to me.

PR #2621, opened 2023-12-14
Update Chinese translations

CookiePieWw:
Ref: https://github.com/newsboat/newsboat/pull/2616#issuecomment-1850666683

Add translation for the newly added string.

I'm working on style consistency already. After that, I plan to have a complete look back at previous translations and then handle the remaining unresolved translations.


der-lyse (approved):
Looks good to me, thanks!

Issue #2625, opened 2023-12-16
`feedlist-title-format` default value when added to config, doesn't  show the same result


lamyergeier:
**Newsboat version (copy the output of `newsboat -v` or the first line of `git show`)**:
Newsboat 2.33.0 

Config file (copy from ~/.newsboat/config or ~/.config/newsboat/config):

```
# Refer:
# 0. [voidrice/config at archi3 · LukeSmithxyz/voidrice](https://github.com/LukeSmithxyz/voidrice/blob/archi3/.config/newsboat/config)
# 1. [Newsboat + Bitchute + mpv + youtube-dl](https://megamorphf.github.io/software/tui/video/2018/08/25/newsboat-bitchute-mpv-youtubedl.html)
# 2. [The Newsboat RSS Feedreader](https://newsboat.org/releases/2.33/docs/newsboat.html)
# 3. [~/.newsbeuter/config](http://moparx.com/configs/newsbeuter/)
# 4. [Newsboat - ArchWiki](https://wiki.archlinux.org/index.php/Newsboat#Configuration)
# 5. [How to open newsboat feed in browser - Ask Ubuntu](https://askubuntu.com/questions/1256425/how-to-open-newsboat-feed-in-browser)

# ----------------------------- Advanced Features -------------------------------------------

# Bookmark: Refer: [Bookmark: Newsboat](https://newsboat.org/releases/2.33/docs/newsboat.html#_bookmarking)
# ========
bookmark-autopilot yes
bookmark-interactive no
bookmark-cmd "~/opt/bin/NewsboatBookmark"

# Filter: Refer: [Filter, Newsboat](https://newsboat.org/releases/2.33/docs/newsboat.html#_filter_language)
# ======
define-filter "minimal-unread" "unread_count >= 2"
run-on-startup select-filter "minimal-unread"

# articles ignore: Refer: [Killfiles, Newsboat](https://newsboat.org/releases/2.33/docs/newsboat.html#_killfiles)
# ======================
ignore-article "*" "title =~ \"Sponsor|Advertisement|[sS]horts?\""
ignore-article "*" "title =~ \"अ|आ|इ|ई|उ\""
ignore-article "*" "title =~ \"Modi|Narendra Modi|bjp|Amit Shah|Rahul Gandhi|Sonia Gandhi|Mamata Banerjee|Adani|Ambani\""
ignore-article "*" "title =~ \"Trump|Biden|Trudeau|Canada\""
[...]

PR #2628, opened 2023-12-16
Update overview of title formats

dennisschagt:
The documentation specifies the title format defaults in two different locations.
One of those locations was missed when removing backticks from those formats in https://github.com/newsboat/newsboat/pull/1572.
Resolves https://github.com/newsboat/newsboat/issues/2625


der-lyse (approved):
Perfect, ta!

PR #2642, opened 2023-12-28
Unify the style of Chinese translations

CookiePieWw:
Most changes are about spaces and punctuation marks.

### DIFF
--- a/po/zh.po
+++ b/po/zh.po
@@ -3,7 +3,7 @@ msgstr ""
 "Project-Id-Version: newsboat 0.7\n"
 "Report-Msgid-Bugs-To: https://github.com/newsboat/newsboat/issues\n"
 "POT-Creation-Date: 2023-12-11 21:31+0300\n"
-"PO-Revision-Date: 2023-12-14 16:00+0800\n"
+"PO-Revision-Date: 2023-12-28 15:00+0800\n"
 "Last-Translator: Yuhan Wang <cookiepieqaq@gmail.com>\n"
 "Language-Team: josh yu <joshyupeng@gmail.com>, Yuhan Wang "
 "<cookiepieqaq@gmail.com>\n"
@@ -22,16 +22,16 @@ msgid ""
 "[-h]\n"
 msgstr ""
 "%s %s\n"
-"用法: %s [-i <file>|-e] [-u <urlfile>] [-c <cachefile>] [-x <command> ...] [-"
+"用法：%s [-i <file>|-e] [-u <urlfile>] [-c <cachefile>] [-x <command> ...] [-"
 "h]\n"
 
 #: newsboat.cpp:47
 msgid "export OPML feed to stdout"
-msgstr "导出OPML种子列表到标准输出"
+msgstr "导出 OPML 种子列表到标准输出"
 
 #: newsboat.cpp:48
 msgid "export OPML 2.0 feed including tags to stdout"
-msgstr "导出OPML2.0种子列表（包括标签）到标准输出"
+msgstr "导出 OPML 2.0 种子列表（包括标签）到标准输出"
 
 #: newsboat.cpp:49
 msgid "refresh feeds on start"
@@ -39,11 +39,11 @@ msgstr "在本软件启动之初刷新种子列表"
 
 #: newsboat.cpp:50 newsboat.cpp:94 newsboat.cpp:100
 msgid "<file>"
-msgstr "文件"
+msgstr "<文件>"
 
 #: newsboat.cpp:50
 msgid "import OPML file"
-msgstr "导入OPML文件"
+msgstr "导入 OPML 文件"
 
 #: newsboat.cpp:54
 msgid "<urlfile>"
@@ -51,11 +51,11 @@ msgstr "<超链文件>"
 
 #: newsboat.cpp:55
 msgid "read RSS feed URLs from <urlfile>"
-msgstr "从超链文件里读取RSS种子列表"
+msgstr "从<超链文件>里读取 RSS 种子列表"
 
 #: newsboat.cpp:60
 msgid "<cachefile>"
-msgstr "缓存文件"
+msgstr "<缓存文件>"
 
 #: newsboat.cpp:61
 msgid "use <cachefile> as cache file"
@@ -75,7 +75,7 @@ msgstr "压缩缓存"
 
 #: newsboat.cpp:73
 msgid "<command>..."
-msgstr "<命令> ..."
+msgstr "<命令>......"
 
 #: newsboat.cpp:74
 msgid "execute list of commands"
@@ -127,7 +127,7 @@ msgstr "清除缓存中未被使用的项"
 
 #: newsboat.cpp:129
 msgid "Files:"
-msgstr "文件: "
+msgstr "文件："
 
 #. i18n: This is printed out by --help before the path to the config file
 #: newsboat.cpp:131
@@ -162,7 +162,7 @@ msgid ""
 "Newsboat is free software licensed under the MIT License. (Type `%s -vv' to "
 "see the full text.)"
 msgstr ""
-"Newsboat 是一个自由软件，根据 MIT 许可证授权。（输入 `%s -vv' 以查看许可证全"
+"Newsboat 是一个自由软件，根据 MIT 许可证授权。（输入`%s -vv'以查看许可证全"
 "文。）"
 
 #: newsboat.cpp:178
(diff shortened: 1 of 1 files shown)


po/zh.po
```diff
@@ -75,7 +75,7 @@ msgstr "压缩缓存"
 
 #: newsboat.cpp:73
 msgid "<command>..."
-msgstr "<命令> ..."
+msgstr "<命令>......"
```
der-lyse, line 78: I reckon some global replacement has gone wild here, ellipsis should only be three periods. Or does Chinese really use six?

po/zh.po
```diff
@@ -271,7 +271,7 @@ msgstr "%N %V - 文章 '%T'（%u 未读, 共有 %t 篇）"
 msgid ""
 "%N %V - Search results for '%s' (%u unread, %t total)%?F? matching filter "
 "'%F'&?"
-msgstr "%N %V - '%s' 的查询结果 (%u 未读, 共有 %t)%?F? 匹配过滤器'%F'&?"
+msgstr "%N %V - '%s' 的查询结果（%u 未读, 共有 %t）%?F? 匹配过滤器'%F'&?"
```
der-lyse, line 274: No idea if this is just a bad font selection on my end but it appears that the closing parenthesis is a mix of a parenthesis combined with a space at the end. I even have no clue if such a thing actually exists or not. In case there is a version without the space at the end, that would be the better option here. Otherwise there will be two spaces if a filter expression exists. Feel free to close this if I'm talking rubbish here. :-)

Righto, after looking through all the changes and seeing the same with other punctuation, I reckon that's just the way it is with either my font or how Chinese is typically rendered. Most likely all good, but I leave the comment just in case. You can enlighten us for sure. :-)

po/zh.po
```diff
@@ -328,51 +328,51 @@ msgstr "%s 第 %u 行"
...
 #, c-format
 msgid "Error while processing command `%s' (%s): %s"
-msgstr "当处理命令`%s'(%s)时出错：%s"
+msgstr "当处理命令`%s' (%s)时出错：%s"
 
 #: src/controller.cpp:154 src/pbcontroller.cpp:233
 #, c-format
 msgid "Starting %s %s..."
-msgstr "启动 %s %s..."
+msgstr "启动 %s %s......"
```
der-lyse, line 341: Double ellipses again. And heaps more down below (I'm not gonna comment on all of them).

po/zh.po
```diff
@@ -929,42 +929,42 @@ msgstr "没有找到操作"
...
 #: src/formaction.cpp:561
 msgid "Feed title: "
-msgstr "种子标题: "
+msgstr "种子标题："
 
 #: src/formaction.cpp:576
 #, fuzzy
 msgid "Saving bookmark on autopilot..."
-msgstr "保存书签..."
+msgstr "保存书签......"
```
der-lyse, line 961: Still fuzzy?

po/zh.po
```diff
@@ -1882,20 +1882,20 @@ msgstr ""
...
 #: src/view.cpp:986
 msgid "Updating query feed..."
-msgstr "更新查询种子..."
+msgstr "更新查询种子......"
 
 #: src/view.cpp:999
 #, fuzzy, c-format
 msgid "Error: couldn't prepare query feed: %s"
-msgstr "错误：无法将种子标记为已读: %s"
+msgstr "错误：无法将种子标记为已读：%s"
```
der-lyse, line 1898: Still fuzzy?

po/zh.po
```diff
@@ -481,37 +481,37 @@ msgstr ""
...
 #: src/controller.cpp:479 src/controller.cpp:564
 msgid "Cleaning up cache..."
-msgstr "清空缓存..."
+msgstr "清空缓存......"
 
 #: src/controller.cpp:491
 #, fuzzy
 msgid "Prepopulating query feeds..."
-msgstr "更新查询种子..."
+msgstr "更新查询种子......"
```
der-lyse, line 506: Is this still fuzzy?

po/zh.po
```diff
@@ -929,42 +929,42 @@ msgstr "没有找到操作"
 #: src/formaction.cpp:333
 #, fuzzy, c-format
 msgid "Not a command: %s"
-msgstr "未知的命令 `%s' "
+msgstr "未知的命令：%s"
```
der-lyse, line 932: Still fuzzy?

der-lyse (commented):
And there it continues, awesome, thanks! I basically just found a potential issue with the ellipsis and some maybe outdated fuzzy markers.

po/zh.po
Minoru: I think we can remove the space at the start of the description of the filter:

```suggestion
msgstr "%N %V - '%s' 的查询结果（%u 未读, 共有 %t）%?F?匹配过滤器'%F'&?"
```

Then it'd look nice no matter if the filter is defined or not.

po/zh.po
```diff
@@ -537,22 +537,22 @@ msgstr ""
 #: src/controller.cpp:719 src/itemutils.cpp:37
 #, c-format
 msgid "Failed to open queue file: %s."
-msgstr "当打开队列文件时出错: %s"
+msgstr "当打开队列文件时出错：%s"
```
Minoru, line 540: There should be a period at the end, too. (I love how Chinese has a special character for that, that little circle looks way cooler than the dot used by the Western languages.)

po/zh.po
```diff
@@ -1336,11 +1336,11 @@ msgstr "在浏览器中打开已选种子中的所有未读项并标记为已读
 
 #: src/keymap.cpp:148
 msgid "Open URL of article, feed, or entry in URL view"
-msgstr "在链接视图(URL view)中打开文章、种子或某一条目(entry)的链接。"
+msgstr "在链接视图（URL view）中打开文章、种子或某一条目（entry）的链接。"
```
Minoru, line 1339:
```suggestion
msgstr "在链接视图（URL view）中打开文章、种子或某一条目（entry）的链接"
```

po/zh.po
```diff
@@ -1336,11 +1336,11 @@ msgstr "在浏览器中打开已选种子中的所有未读项并标记为已读
...
 
 #: src/keymap.cpp:155
 msgid "Open URL of article, feed, or entry in a browser, non-interactively"
-msgstr "在浏览器中打开文章、种子或某一条目(entry)的链接，非交互式。"
+msgstr "在浏览器中打开文章、种子或某一条目（entry）的链接，非交互式。"
```
Minoru, line 1343:
```suggestion
msgstr "在浏览器中打开文章、种子或某一条目（entry）的链接，非交互式"
```

po/zh.po
CookiePieWw: Yes, Chinese uses `......` as the ellipsis. Besides, there's indeed a problem about ellipsis.
https://github.com/newsboat/newsboat/blob/c882fcfa721da02b17a19d3f5d173729f3ec42bc/po/zh.po#L78
This phrase is displayed by `newsboat --help` and not treated as a complete sentences. I initially replaced all of the ellipsis to Chinese ones, but it seems too long and not consistent with its surroundings (typically some command line arguments).
Another example which may be more comprehensive:
https://github.com/newsboat/newsboat/blob/c882fcfa721da02b17a19d3f5d173729f3ec42bc/po/zh.po#L25
I decided to act like this: for command line arguments, use `...` to keep it simple and consistent, for others, use `......` instead, since its truly a Chinese sentence. I replaced some of them manually, but this one remains.

po/zh.po
CookiePieWw: Fuzzy items remain since this PR only focuses on styles. Same as below.

po/zh.po
CookiePieWw: Got it, I'll make some changes later. BTW, ancient Chinese writers in the Song dynasty used to use a bigger `○` to split chapters, middle ones to split paragraphs, and small ones (which we still use today) to split sentences. They really like circles.

po/zh.po
CookiePieWw: After having a close look, I found it interesting that Chinese parenthesis  `（）` seems to have different radian with each other. It may be caused by the renderer.

po/zh.po
der-lyse: Wow, interesting! Quite fascinating to see what other parts of the world do. :-)

po/zh.po
der-lyse: I see.

po/zh.po
der-lyse: Just for reference, on my end it looks like this. I selected the opening parenthesis of your comment above to highlight the parenthesis' left padding.

> parentheses

Quite something to learn here for typography nerds. ;-) I love it!

der-lyse (approved):

po/zh.po
CookiePieWw: Ah I see. The padding is intended. In monospaced font, all Chinese characters occupy twice the width of English characters, so as the punctuation (except the ellipsis, it's quite long, some times we even use `。。。。。。` in hand-writing). A similar pair to parenthesis is `《》` which we use for titles of books. They also have left and right paddings.

PR #2648, opened 2024-01-02
Split key bindings and events into a base key with some modifiers (shift, control, alt)

dennisschagt:
STFL events currently are regular strings containing a key name, optionally prefixed with `^` (for the "control" modifier) or capitalized (for the "shift" modifier).
I expect a replacement for STFL might contain this information in a more structured way (a separate string for the key name and some flags to indicate active modifiers, e,g,: [Crossterm's `KeyEvent`](https://docs.rs/crossterm/latest/crossterm/event/struct.KeyEvent.html)).

I expect this PR makes it easier to move away from STFL and, at the time, easier to introduce the new `bind` option.

### DIFF
--- a/include/keycombination.h
+++ b/include/keycombination.h
@@ -1,0 +1,33 @@
+#ifndef NEWSBOAT_KEYCOMBINATION_H_
+#define NEWSBOAT_KEYCOMBINATION_H_
+
+#include <ostream>
+#include <string>
+
+namespace newsboat {
+
+class KeyCombination {
+public:
+	KeyCombination(const std::string& key, bool shift, bool control, bool alt);
+
+	static KeyCombination from_bindkey(const std::string& input);
+	std::string to_bindkey_string() const;
+
+	bool operator==(const KeyCombination& other) const;
+	bool operator<(const KeyCombination& rhs) const;
+
+	std::string get_key() const;
+	bool has_shift() const;
+	bool has_control() const;
+	bool has_alt() const;
+
+private:
+	std::string key;
+	bool shift;
+	bool control;
+	bool alt;
+};
+
+} // namespace newsboat
+
+#endif /* NEWSBOAT_KEYCOMBINATION_H_ */
--- a/include/keycombination.h
+++ b/include/keycombination.h
@@ -6,10 +6,27 @@
 
 namespace newsboat {
 
+enum class ShiftState {
+	Shift,
+	NoShift,
+};
+
+enum class ControlState {
+	Control,
+	NoControl,
+};
+
+enum class AltState {
+	Alt,
+	NoAlt,
+};
+
 class KeyCombination {
 public:
-	explicit KeyCombination(const std::string& key, bool shift = false, bool control = false,
-		bool alt = false);
+	explicit KeyCombination(const std::string& key,
+		ShiftState shift = ShiftState::NoShift,
+		ControlState control = ControlState::NoControl,
+		AltState alt = AltState::NoAlt);
 
 	static KeyCombination from_bindkey(const std::string& input);
 	std::string to_bindkey_string() const;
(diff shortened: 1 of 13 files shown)


include/keycombination.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include <ostream>
+#include <string>
+
+namespace newsboat {
+
+class KeyCombination {
+public:
+	explicit KeyCombination(const std::string& key, bool shift = false, bool control = false,
+		bool alt = false);
```
der-lyse, line 12: I just want to point out that these three booleans can be mixed up easily as there is no natural order for <kbd>Shift</kbd>, <kbd>Ctrl</kbd> and <kbd>Alt</kbd> in my opinion. Any order is equally valid.

However, it might be quite some overkill to introduce typed enums for these modifiers, e.g. `NoShift`/`Shift` etc. A single combinable modifier type would be probably even better, there's a reason, that lots of APIs do it like that.

Anyways, maybe not so bad as this constructor is only used at a few places. Just tests might still benefit from dedicated types.

src/keycombination.cpp
```diff
@@ -0,0 +1,82 @@
...
+	std::string key = input;
+	bool shift = false;
+	bool control = false;
+	bool alt = false;
+
+	if (key.length() == 1 && std::isupper(key[0])) {
+		shift = true;
+		key = std::tolower(key[0]);
+	}
+	if (key.length() == 2 && key[0] == '^') {
```
der-lyse, line 27: This can be an `else if`.

src/keycombination.cpp
```diff
@@ -0,0 +1,82 @@
...
+std::string KeyCombination::to_bindkey_string() const
+{
+	std::string output = key;
+	if (control && output.length() == 1) {
+		output = std::string("^") + static_cast<char>(std::toupper(output[0]));
+	}
+	if (shift && output.length() == 1) {
+		output = std::toupper(output[0]);
+	}
+	return output;
```
der-lyse, line 43: Personally, I'm a big fan of early returns, not sure if this is also a thing in Newsboat's code base. If not, I'd use an `else if` here, too.

src/keycombination.cpp
```diff
@@ -0,0 +1,82 @@
...
+	return control;
+}
+
+bool KeyCombination::has_alt() const
+{
+	return alt;
+}
+
+
+
```
der-lyse, line 81: Quite a bunch of empty lines. :-)

dennisschagt:
Thanks, good points, applied

include/keycombination.h
dennisschagt: I've changed these into separate enums.
Not sure if it is easer to read, but it is easier to avoid accidental issues in reordering the arguments.

der-lyse: Yes, the `FooState::` clutters the code quite a bit, that's right. But in the end it is hopefully faster to read through that lengthy code instead of jumping back to the definition all the time to see the argument order. *Maybe* in tests some `using enum …` helps in that regard, not sure, I don't consider myself a real C++ programmer.

In `to_bindkey_string()` the use of the boolean modifier helper methods *might* make the code slightly shorter and hence a tad more readable. But that's the only other thing I can think of.

Minoru (approved):
Great idea, thanks for implementing this!

I notice some `KeyMap` methods weren't converted, and still take/return strings — do you plan to convert them later? I'd be nice to use this class throughout the entirety of Newsboat. Converting `OpDesc` will be a chore I guess, since that will entail converting the giant list of default keys in keymap.cpp.

dennisschagt:
Thanks both for the reviews!

> I notice some KeyMap methods weren't converted, and still take/return strings — do you plan to convert them later?

Yes, I plan to convert those when implementing the key-mapping storage for the new bind command

> Converting OpDesc will be a chore I guess, since that will entail converting the giant list of default keys in keymap.cpp.

I didn't think about converting that one yet, but I might pick that up soon

PR #2652, opened 2024-01-05
Move utils bridged methods into their own ffi module

dennisschagt:
Discovered that we can use `Pin<&mut ...>` to allow mutation of Rust types defined in another cxx bridge module, based on:
- https://cxx.rs/extern-c++.html#opaque-c-types
  > Unlike extern Rust types and shared types, an extern C++ type is not
  > permitted to be passed by plain mutable reference &mut MyType across the
  > FFI bridge. For mutation support, the bridge is required to use Pin<&mut
  > MyType>. This is to safeguard against things like mem::swap-ing the
  > contents of two mutable references, given that Rust doesn't have
  > information about the size of the underlying object and couldn't invoke
  > an appropriate C++ move constructor anyway.
- https://github.com/dtolnay/cxx/issues/942

I'm already happy with this result (took me long enough to try a lot of different options which did not work).
If/when https://github.com/dtolnay/cxx/issues/353 is resolved, we might be able to (re)use these types again within boxes and return them (instead of using the mutable references).

### DIFF
--- a/include/filepath.h
+++ b/include/filepath.h
@@ -36,7 +36,7 @@ public:
 	{
 	}
 
-	Filepath(rust::Box<filepath::bridged::PathBuf> rs_object)
+	Filepath(rust::Box<filepath::bridged::PathBuf>&& rs_object)
 		: rs_object(std::move(rs_object))
 	{
 	}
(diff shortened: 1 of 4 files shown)


Minoru (approved):
Quite clever! It's nice that we can keep the functions in their respective modules, even if that requires a bit of "output parameter" ugliness behind the scenes.

PR #2653, opened 2024-01-06
Use PathBuf on FFI (replacing `&str` and `String`)

dennisschagt:
Resolves https://github.com/newsboat/newsboat/issues/2644

There are a few places where we now convert `Filepath` into `std::string` to then directly convert into a `Filepath` again.
I decided to postpone work on that given that we will find them already when working on https://github.com/newsboat/newsboat/issues/2326#issuecomment-1872314543

### DIFF
--- a/include/cliargsparser.h
+++ b/include/cliargsparser.h
@@ -72,11 +72,11 @@ public:
 
 	nonstd::optional<Filepath> config_file() const;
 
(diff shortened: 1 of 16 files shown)


Minoru (approved):
Excellent, thank you very much Dennis!

PR #2660, opened 2024-01-11
Add utilities for detecting text encoding based on byte-order-mark and XML declaration

dennisschagt:
These commits introduce a few utility functions for detecting the character encoding of text.
This can be used in a future PR to resolve several of our encoding issues (probably extending or obsoleting https://github.com/newsboat/newsboat/pull/2340).

### DIFF
--- a/rust/libnewsboat/src/charencoding.rs
+++ b/rust/libnewsboat/src/charencoding.rs
@@ -1,3 +1,13 @@
+use nom::branch::alt;
+use nom::bytes::complete::tag;
+use nom::character::complete::{alpha1, alphanumeric1, digit1, space0};
+use nom::combinator::recognize;
+use nom::multi::many0;
+use nom::sequence::pair;
+use nom::IResult;
+use std::str;
+
+/// Returns charset if it can be derived from the optional byte-order-mark
 pub fn charset_from_bom(content: &[u8]) -> Option<&'static str> {
     match content {
         [0x00, 0x00, 0xFE, 0xFF, ..] => Some("UTF-32BE"),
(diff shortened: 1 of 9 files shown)


rust/libnewsboat/src/charencoding.rs
```diff
@@ -0,0 +1,137 @@
...
+        [0xEF, 0xBB, 0xBF, ..] => Some("UTF-8"),
+        [0x2B, 0x2F, 0x76, ..] => Some("UTF-7"),
+        _ => None,
+    }
+}
+
+/// Returns charset if the encoding is specified in an XML declaration
+///
+/// Example XML declaration: `<?xml version="1.0" encoding="UTF-8"?>`
+pub fn charset_from_xml_declaration(content: &[u8]) -> Option<String> {
```
dennisschagt, line 26: It might be useful to also implement the idea from the "Without a Byte Order Mark" table in https://www.w3.org/TR/REC-xml/#sec-guessing-no-ext-info (I believe libxml2 also does that, so might even be required to avoid regressions)

dennisschagt: I've implemented this for UTF-16 and UTF-32 but did not implement the detection for the specific EBCDIC variant.
This would cause regressions if we do all of the encoding conversions in Newsboat, but I see an alternative; some untested/pseudo code:
```
void get_feed(uri) {
    auto [content, content_type_charset] = download(uri); // optional charset from content-type http header
    auto bom_charset = charset_from_bom(content);
    auto xml_declaration_charset = charset_from_xml_declaration(content);

    optional<string> charset = {};
    if (charset_from_bom.has_value()) {
        charset = charset_from_bom.value();
    } else if (xml_declaration_charset.has_value()) {
        charset = = xml_declaration_charset.value();
    } else if (content_type_charset.has_value()) {
        charset = = content_type_charset.value();
    }

    if (charset.has_value()) {
        auto text = convert_text(charset.value(), "utf-8");
        return parseXml(text, "utf-8"); // Force libxml2 to use utf-8 and make it ignore the xml declaration
    } else {
        return parseXml(content, NULL); // Use libxml2's charset auto-detection (what we are currently doing)
    }
}
```

dennisschagt:
Merging this as it cannot break anything (new functionality, not yet connected) and I would like to continue building some functionality on top of these changes.

PR #2670, opened 2024-02-01
adds simple wallabag bookmark plugin

pdxmph:
A super simple bookmark plugin for Wallabag. Could stand to be refactored to use Ruby's native net:http library to remove the dependency on httparty. This one only sends the URL since Wallabag does its own extraction of title and description. 

README has step-by-step on how to configure and deploy.


dennisschagt:
Nice work, thanks!

PR #2688, opened 2024-02-24
Add a filter to convert twtxt feeds

cicxjo:
Twtxt is a small plain-text microblogging protocol.

Twtxt2rss converts twtxt feeds into RSS feeds and enables people to follow twtxt feeds without a dedicated client.


dennisschagt (approved):
Thanks, looks good

PR #2711, opened 2024-03-20
Add xslt transformation for replacing Slashdot's newlines

dennisschagt:
User **lv8pv** mentioned on the \#newsboat IRC channel that the newlines in Slashdot articles are not shown in Newsboat.
Feed url: http://rss.slashdot.org/Slashdot/slashdotMain

I don't know if newsboat does anything wrong in parsing the feed, but this script can be used to transform newlines into explicit html `<br>` elements.

### DIFF
--- a/contrib/slashdot-replace-newlines.xslt
+++ b/contrib/slashdot-replace-newlines.xslt
@@ -1,0 +1,24 @@
+<!--
+Replaces newline characters in description field of Slashdot's RDF/RSS feed with html linebreaks (`<br>`)
+Uses xsltproc (part of libxslt)
+
+Can be used with a filter url, for example:
+"filter:xsltproc /usr/share/doc/newsboat/contrib/slashdot-replace-newlines.xslt -:https://rss.slashdot.org/Slashdot/slashdotMain"
+-->
+<xsl:stylesheet version="1.0"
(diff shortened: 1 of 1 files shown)


contrib/slashdot-replace-newlines.xslt
```diff
@@ -0,0 +1,24 @@
...
+    <xsl:template match="node()|@*">
+        <xsl:copy>
+            <xsl:apply-templates select="node()|@*"/>
+        </xsl:copy>
+    </xsl:template>
+
+    <xsl:template match="purl:description">
+        <xsl:copy>
+            <!-- '&#10;' encodes a newline character (\n) -->
+            <xsl:value-of select="str:replace(node(), '&#10;', '&lt;br&gt;')"/>
```
dennisschagt, line 21: I had some trouble getting the `replace()` call to work.
Finally found a solution at https://stackoverflow.com/questions/3067113/xslt-string-replace/59815107#59815107
However, a comment below it mentions that it is (or was) not available on the macOS 10.13 version of xsltproc.

der-lyse (approved):
Luckily, I did not have to work with XSLT for many years now. But I reckon this looks alright. 8-)

dennisschagt:
Thanks!

PR #2718, opened 2024-03-23
Swedish translation: new strings and updates

deob83:
Adding translation of new strings and tweaking translation of previously translated strings.


der-lyse (approved):
Looks good to me, thank you!

PR #2726, opened 2024-03-31
Cache: Support multiple enclosures per item

dennisschagt:
These changes update the database scheme to support multiple enclosures per item.
This is a preparation for implementing https://github.com/newsboat/newsboat/issues/2638.


src/cache.cpp
```diff
@@ -397,6 +430,32 @@ static const schema_patches schemaPatches{
...
+			"INSERT INTO enclosure (item_id, url, type, description, description_mime_type, enqueued)"
+			" SELECT id, enclosure_url, enclosure_type, enclosure_description, enclosure_description_mime_type, enqueued"
+			" FROM rss_item"
+			" WHERE enclosure_url IS NOT NULL AND enclosure_url != '';",
+
+			"ALTER TABLE rss_item DROP COLUMN enclosure_url;",
+			"ALTER TABLE rss_item DROP COLUMN enclosure_type;",
+			"ALTER TABLE rss_item DROP COLUMN enclosure_description;",
+			"ALTER TABLE rss_item DROP COLUMN enclosure_description_mime_type;",
+			"ALTER TABLE rss_item DROP COLUMN enqueued;",
```
dennisschagt, line 456: When removing these columns, old versions of newsboat will fail to load articles from the database due to failing queries.

@Minoru I think you have mentioned previously that you want to support the case where people open the cache file with a newer version of newsboat and afterwards try to use it from an older version again (possibly from a different system).
If you still feel strongly about that, I'm fine with removing the above 5 `DROP COLUMN` lines from this PR.
However, in that case I would like to prepare a changeset which reintroduces the `DROP COLUMN` lines and possibly a merging of enclosure data from the `rss_item` table into the `enclosure` table again. (and I think it would be nice in that case to make the extend of this forward? compatiblity a bit more explicit, e.g. only supporting this for a maximum of 2 or 3 minor releases)

dennisschagt: > ... support the case where people open the cache file with a newer version of newsboat and afterwards try to use it from an older version again

I guess this might also be very useful in keeping git bisect easy to use, so I've commented out the cleanup steps.

PR #2731, opened 2024-04-07
Update testcoverage build dependencies

dennisschagt:
Current rust nightly seems to be based on LLVM 18:
```
$ rustc +nightly --version --verbose
rustc 1.79.0-nightly (aa1c45908 2024-04-06)
binary: rustc
commit-hash: aa1c45908df252a5b0c14e1bcb38c6c55ae02efe
commit-date: 2024-04-06
host: x86_64-unknown-linux-gnu
release: 1.79.0-nightly
LLVM version: 18.1.2
```

### DIFF
--- a/.github/workflows/coveralls.yaml
+++ b/.github/workflows/coveralls.yaml
@@ -5,13 +5,13 @@ on: [push, pull_request]
 jobs:
   gen_coverage:
     name: Calculate test coverage
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     env:
-      # We use Rust Nightly, which builds upon LLVM 12. To ensure best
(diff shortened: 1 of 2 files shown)


.github/workflows/coveralls.yaml
```diff
@@ -5,13 +5,13 @@ on: [push, pull_request]
 jobs:
   gen_coverage:
     name: Calculate test coverage
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
dennisschagt, line 8: I tried to update to ubuntu-23-10, but [that is apparently not supported](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)

PR #2737, opened 2024-04-15
Fix default color documentation

flyxyz123:
Fix default color documentation according to current code ./stfl/feedlist.stfl

Commit 368048b changed listnormal_unread and listfocus_unread, then commit a8971bd fixed those, but they did not correct the documentation. Commit 602d35b added hint-keys-delimiter, hint-separator, and hint-description; then commit 63b6c2f added the wrong documentation.


dennisschagt (approved):
Nice, thanks!

PR #2697, opened 2024-03-09
Update to chrono 0.4.35

dennisschagt:
Supersedes https://github.com/newsboat/newsboat/pull/2696 (this PR fixes warnings caused by the update)

PR #2747, opened 2024-05-01
Require chrono 4.34 since Duration was renamed to TimeDelta

tsipinakis:
I'm not sure why an incompatible change was made in a minor release...

Edit: Or is it actually compatible just that we used the new features without restricting the minor release?


dennisschagt:
> Or is it actually compatible just that we used the new features without restricting the minor release?

Indeed, might have been caused by me in https://github.com/newsboat/newsboat/pull/2697
Anyway, this change looks good to me.

PR #2753, opened 2024-05-18
Cleanup unused overloaded methods

dennisschagt:
I noticed that our production code always instantiates `FeedRetriever` with a non-null `CurlHandle`.
Changed the argument to a reference to make this obvious (and updated tests to make them use `FeedRetriever` in the same way as the production code).
Extended these changes to external APIs and other places where we use a `CurlHandle`.

Additionally I noticed that my C++ Vim plugin started highlighting unused header includes.
~~I've removed those unnecessary header includes in the files I had open for this PR.
If preferred, I can take out those changes and apply them in a separate PR.~~
-> I ran into some issues so will move these changes into a separate PR

### DIFF
--- a/test/rssfeed.cpp
+++ b/test/rssfeed.cpp
@@ -3,6 +3,7 @@
 #include "3rd-party/catch.hpp"
 #include "cache.h"
 #include "configcontainer.h"
+#include "curlhandle.h"
 #include "feedretriever.h"
 #include "rssparser.h"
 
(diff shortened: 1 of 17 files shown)


test/rssfeed.cpp
```diff
@@ -3,11 +3,11 @@
 #include "3rd-party/catch.hpp"
 #include "cache.h"
 #include "configcontainer.h"
+#include "curlhandle.h"
 #include "feedretriever.h"
 #include "rssparser.h"
 
 #include "test_helpers/envvar.h"
-#include "test_helpers/stringmaker/optional.h"
```
dennisschagt, line 10: Might need to add this again.
The Vim plugin I'm using for C++ highlights this because it is not directly used.
However, it is used indirectly by Catch2 to convert `nonstd::optional<...>` types to string.

dennisschagt: I will update in a bit.
I have to rebase anyway when https://github.com/newsboat/newsboat/pull/2755 is merged (to resolve build issue)

PR #2792, opened 2024-06-29
Readeck bookmark scripts

rainbow-bamboo:
Hi,
This adds two scripts to allow bookmarking to your [Readeck](https://readeck.org) installation.

They are Clojure scripts which need the added dependency of [Babashka](https://github.com/babashka/babashka#installation) to run, but using babashka allows for much faster startup times and memory usage than would be possible with JVM Clojure.

See the readme for more details :)


dennisschagt (approved):
Nice, I didn't try running it myself but both code and readme look good to me!

PR #2801, opened 2024-07-20
Test handling of character encodings specified in xml encoding attribute and http header

dennisschagt:
This splits of part of the work from https://github.com/newsboat/newsboat/pull/2340

### DIFF
--- a/test/rsspp_parser.cpp
+++ b/test/rsspp_parser.cpp
@@ -1,8 +1,15 @@
 #include "rss/parser.h"
 
+#include <cstdint>
+
 #include "3rd-party/catch.hpp"
+#include "curlhandle.h"
 #include "rss/exception.h"
+#include "strprintf.h"
 #include "test_helpers/exceptionwithmsg.h"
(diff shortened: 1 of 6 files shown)


test/rsspp_parser.cpp
```diff
@@ -378,3 +385,120 @@ TEST_CASE("Feed authors and source authors in atom feed", "[rsspp::Parser]")
...
+
+	rsspp::Parser parser;
+	CurlHandle easyhandle;
+	auto parsed_feed = parser.parse_url(url, easyhandle);
+
+	REQUIRE(parsed_feed.items.size() == 1);
+	REQUIRE(parsed_feed.title == title_utf8);
+}
+
+TEST_CASE("parse_url() only converts once even when encoding specified twice (xml encoding and http header)", "[rsspp::Parser]")
```
dennisschagt, line 454: These last 2 tests might seem a bit useless now as our implementation does not yet look at the `charset=...` part of `Content-Type` headers.
However, I expect to introduce that soon and want to use these tests to avoid regressions.

Issue #2791, opened 2024-06-28
[ Feature request]: Show error message in the status line if one of the feed fails to retreive.


ZaxonXP:
I the current version of newsboat there is no way to know if one of the feed failed to download. The only way to discover that there is something wrong is by noticing the absence of the new posts, which is not very obvious for feeds that contains less frequent posts.

It would be nice if newsboat could:
1) show the error message when one or more feeds failed to download new posts
2) create entry in some error log, so user can check which feeds failed.

Can you please implement such features in the one of the next releases?

Kind regards,
Piotr

juntuu:
I've also wanted a better view & visibility for download errors, and the issue #2765 has quite similar goal (it would play well with option 1.).

There are three ways I know of, that can show this information currently:

1. The feedlist has [format option S](https://newsboat.org/releases/2.31/docs/newsboat.html#feedlist-format-S) for download status. I put this at the start of the line, so I see which feeds failed to load. This does not show the reason.
2. Logging gives a message with a reason. The [error-log option](https://newsboat.org/releases/2.31/docs/newsboat.html#error-log) is good for this, but other logging options also work (but have more unrelated messages).
3. The logged error message is also briefly shown in the UI, but there's only room for one message and it gets quickly overwritten (e.g. by the reload progress messages). So this is not practical.

ZaxonXP:
@juntuu: Looks like option 1 is the one that can do for now. :)

Thanks for pointing out this feature.

ZaxonXP:
Maybe additional error log window and on the status line a message like:
```Some errors occured. Press <define_key> to see the log```

juntuu:
I had similar idea about error log window inside the application.

I implemented a demo in https://github.com/juntuu/newsboat/tree/message-view
- UserError level log messages are kept in memory, and shown in the view as a simple list
- The view is adapted from help dialog (I don't really know stfl or the Newsboat's UI logic)
- The view is bound to `m` and should work from most other views

This does not have any indication about new messages in status bar or similar. It could also be good to allow clearing the messages (other than restarting the program).

---

Another idea I had is adding the number of failed reloads to the notification format. I don't use the notification system myself, but this would make sense to me. Demo implementation here https://github.com/juntuu/newsboat/tree/notify-on-failed-reloads
- Failed reloads are counted
- The count is exposed as `E`(rrors) format identifier
- `E` is empty if there are no errors, so that this works with conditional format
- Notification is shown if any errors happened

---

[...]

PR #2807, opened 2024-07-23
Add download_status attribute for feeds

dennisschagt:
Just a quick experiment to see if this works for https://github.com/newsboat/newsboat/issues/2791#issuecomment-2213805073
This is not a full solution to https://github.com/newsboat/newsboat/issues/2791

Example config option (press `f` to select the defined filter):
```
define-filter "feeds - unread and/or error" "unread_count > 0 or download_status = \"x\""
```

This filter does work but it might break easy usage of `toggle-show-read-feeds` (has no visible effect as long as the filter is active)


dennisschagt:
There might be better ways to resolve the linked issue so I will close this for now

PR #2849, opened 2024-09-09
chore(l10n): Update Polish translation before 2.37 release

carno:
Ref: https://github.com/newsboat/newsboat/pull/2847#issuecomment-2338656080


der-lyse (approved):
LGTM

PR #2852, opened 2024-09-11
Update German translations

der-lyse:


dennisschagt (approved):
Looks good, also gave me the right idea for the Dutch translation, so thanks! :)

der-lyse:
Yeah, this one was a bit tricky. I started with "fehlend", but then checked [dict.cc](https://www.dict.cc/?s=missing) for alternatives. Not so many suggestions come up for [Dutch](https://ennl.dict.cc/?s=missing), though.

PR #2859, opened 2024-09-15
Show keymap hint for "clear filter" when filter is active

dennisschagt:
Came up with the idea while looking into https://github.com/newsboat/newsboat/issues/2857.

This PR introduces a new translatable string so merging might need to wait until after the upcoming release.

### DIFF
--- a/include/dialogsformaction.h
+++ b/include/dialogsformaction.h
@@ -12,7 +12,7 @@ public:
 	~DialogsFormAction() override = default;
 	void prepare() override;
 	void init() override;
-	const std::vector<KeyMapHintEntry>& get_keymap_hint() const override;
+	std::vector<KeyMapHintEntry> get_keymap_hint() const override;
 	std::string id() const override
 	{
 		return "dialogs";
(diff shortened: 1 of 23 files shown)


Minoru:
I like the idea!

A bit of bikeshedding: I'd prefer the "Help" hint to always come last, because it's like an ellipsis. How about putting the new hint at the very start?

dennisschagt:
Thanks!
Rebased to make sure we get build results for FreeBSD 14.1.

> A bit of bikeshedding: I'd prefer the "Help" hint to always come last, because it's like an ellipsis. How about putting the new hint at the very start?

Nice idea, done

dennisschagt:
In case it helps, `git -range-diff @{upstream}...` output from just before I force-pushed:
image

dennisschagt:
I guess a similar thing can be done for tags (show `^T:Clear Tag` if a tag is selected)

dennisschagt:
> I guess a similar thing can be done for tags (show ^T:Clear Tag if a tag is selected).
> And in help, I guess we can hide the ^F:Clear by default (and only show it after a search/highlight was executed).

Done.
For now, I've put those in separate commits to avoid a rebase.
If preferred, I'm completely fine with squashing all commits into 1.

PR #2869, opened 2024-09-24
support negative indexing in feed/item list

notdian:
This PR adds support for negative indexing to the feed and item lists, allowing users to quickly access elements from the end of a list. For example, using -1 will retrieve the last item. This improves ease of use and aligns with common indexing practices, making list navigation more intuitive.


dennisschagt:
I just tested this locally.
It seems to work fine if all feeds/items are shown, but it breaks down if the list is filtered (e.g. by pressing lowercase `L` to only show unread items, or when applying a filter/tag)
In those cases, most (or all?) negative numbers result in an error message: "Position not visible!"

The regular, non-negative number handling goes to items based on their number in the complete/unfiltered list.
Those numbers are shown on the left side of the screen and are stable (don't change) if the list is filtered.
You can start typing these numbers, even without pressing `:` (the first number will automatically open the "cmdline".
I'm wondering if there are any cases where the negative indexing would be faster.
Do you have any examples where it is?

Either way, sorry for the late review.
It is always nice to see new contributors but I'm not sure if we will take in this change.

Issue #2870, opened 2024-09-25
Issue with Miniflux config


hyde-stevenson:
`
Newsboat 2.36.0
System: Linux 6.4.0-1-amd64 (x86_64)
Compiler: g++ 14.2.0
ncurses: ncurses 6.4.20240113 (compiled with 6.5)
libcurl: libcurl/8.7.1 GnuTLS/3.8.5 zlib/1.3.1 brotli/1.0.9 zstd/1.5.5 libidn2/2.3.4 libpsl/0.21.2 libssh2/1.11.0 nghttp2/1.55.1 librtmp/2.3 OpenLDAP/2.5.13 (compiled with 8.9.1)
SQLite: 3.42.0 (compiled with 3.46.0)
libxml2: compiled with 2.12.7
Config file (copy from ~/.newsboat/config or ~/.config/newsboat/config):
```

## Config 

`
auto-reload yes
bind-key ^O bookmark
bookmark-autopilot yes
bookmark-cmd "~/bin/addnote"
bookmark-interactive yes
browser "/usr/bin/librewolf -new-tab %u"
feed-sort-order unreadarticlecount
max-items 0
urls-source "miniflux"
miniflux-url "https://miniflux.url.com/"
miniflux-login "hyde"
miniflux-password "XXXXXXXXXXXXXXXXXXXXXXXX"

#miniflux-show-special-feeds "no"
`

Steps to reproduce the issue:

1. If i add the last line in the config, I get : 
`
It looks like you haven't configured any feeds in your Miniflux account. Please do so, and try again.
`

2. If I remove it, newsboat launches, but I only have the "Starred" menu without anything in it. I tried to "star"
one article to see if I would see when I try to reload with the R key. Nothing happens. 
I get just the message : 
`
(1/1) Loading starred ... 
`
It's like it doesn't see  the feeds I have in Miniflux.

**Other info you think is relevant**:

I did a strace on the process, when I hit just r not R, I get this message : 
`
newfstatat(AT_FDCWD, "/home/hyde/.newsboat/cache.db-journal", 0x7ffe5a715fa0, 0) = -1 ENOENT (No such file or directory)
newfstatat(AT_FDCWD, "/home/hyde/.newsboat/cache.db-wal", 0x7ffe5a715fa0, 0) = -1 ENOENT (No such file or directory)
`

I configured Read You, the mobile app, and I can reach, and retrieve all the feeds. So, I dont think it's linked with the minflux installation.

[...]

PR #2876, opened 2024-09-29
Report error when miniflux authentication check fails outside 401

dennisschagt:
Intended to make issues like https://github.com/newsboat/newsboat/issues/2870 easier to diagnose.

I'm not sure why `get_subscribed_urls()` only logs errors without stopping startup.
Only reason I can come up with is that it might also get called after an `edit-urls` command, in which case we probably don't want to crash/stop newsboat.

### DIFF
--- a/src/minifluxapi.cpp
+++ b/src/minifluxapi.cpp
@@ -2,9 +2,11 @@
 
 #include <cinttypes>
 #include <curl/curl.h>
+#include <iostream>
 #include <thread>
 
 #include "3rd-party/json.hpp"
+#include "config.h"
 #include "curlhandle.h"
 #include "logger.h"
 #include "remoteapi.h"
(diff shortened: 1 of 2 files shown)


src/minifluxapi.cpp
```diff
@@ -43,12 +45,21 @@ bool MinifluxApi::authenticate()
...
 	curl_easy_getinfo(handle.ptr(), CURLINFO_RESPONSE_CODE, &response_code);
 
 	if (response_code == 401) {
 		return false;
 	}
+	if (response_code < 200 || response_code > 299) {
+		const auto url = server + path;
+		std::cerr << strprintf::fmt(
+				_("Authentication check using %s failed (HTTP response code: %ld)"),
+				url,
```
dennisschagt, line 59: I like to include this URL because it makes for an easy way to debug a few issues using regular browser or curl.

Current output:
image

Wrong username/password output is not changed:
image

PR #2879, opened 2024-09-30
Properly install shell completions and add zsh completions

dennisschagt:
Taken from the Newsboat mailing list, [submitted by Ludovico Gerardi](https://groups.google.com/g/newsboat/c/zGuycMQYzF0):

> Currently the files for shell completions are stored into the
'contrib' directory, and the build system installs them under $docdir.
However that is not useful for the final user because by default shells
don't use that path when looking for completion files. This patchset
addresses that by installing those files under the right directories. It
also provides completions for zsh.


dennisschagt:
I had to make two small adjustments to make sure the files are also removed when running `make uninstall`.
Range-diff to show my changes to the original diffs:
image

dennisschagt:
Changes look fine to me so, if there are no negative reviews, I plan to merge this in 3 days.

Makefile
```diff
@@ -361,9 +361,9 @@ install-docs: doc
 	$(MKDIR) $(DESTDIR)$(docdir)
 	$(INSTALL) -m 644 doc/xhtml/* $(DESTDIR)$(docdir)
 	$(INSTALL) -m 644 CHANGELOG.md $(DESTDIR)$(docdir)
-	find contrib/ -type d -print0 | xargs -0 -I@ $(MKDIR) $(DESTDIR)$(docdir)/@
-	find contrib/ -type f -perm /0111 -print0 | xargs -0 -I@ install -m 755 @ $(DESTDIR)$(docdir)/@
-	find contrib/ -type f ! -perm /0111 -print0 | xargs -0 -I@ install -m 644 @ $(DESTDIR)$(docdir)/@
+	find contrib/ -type d ! -wholename 'contrib/completions' -print0 | xargs -0 -I@ $(MKDIR) $(DESTDIR)$(docdir)/@
+	find contrib/ -type f ! -wholename 'contrib/completions*' -perm /0111 -print0 | xargs -0 -I@ install -m 755 @ $(DESTDIR)$(docdir)/@
+	find contrib/ -type f ! -wholename 'contrib/completions*' ! -perm /0111 -print0 | xargs -0 -I@ install -m 644 @ $(DESTDIR)$(docdir)/@
```
Minoru, line 366: https://groups.google.com/g/newsboat/c/zGuycMQYzF0/m/RcO8sAHWAAAJ

> `-wholename` appears to be a non-standard synonym for `-path`. Even the manpage notes that it's a less portable alternative: https://www.man7.org/linux/man-pages/man1/find.1.html It seems to be supported by FreeBSD, but is missing from OpenBSD: https://man.openbsd.org/find

```suggestion
	find contrib/ -type d ! -path 'contrib/completions' -print0 | xargs -0 -I@ $(MKDIR) $(DESTDIR)$(docdir)/@
	find contrib/ -type f ! -path 'contrib/completions*' -perm /0111 -print0 | xargs -0 -I@ install -m 755 @ $(DESTDIR)$(docdir)/@
	find contrib/ -type f ! -path 'contrib/completions*' ! -perm /0111 -print0 | xargs -0 -I@ install -m 644 @ $(DESTDIR)$(docdir)/@
```

PR #2882, opened 2024-10-01
fix mkdir invocation in install recipe

humenda:


juntuu:
The `-p` flag is expected to be part of the `$(MKDIR)` variable.

See: https://github.com/newsboat/newsboat/blob/master/Makefile#L115

humenda:
Pardon, my oversight.

juntuu:
No broblem :)

PR #2968, opened 2024-12-30
Docs: use proper markup for key presses

der-lyse:
Back in the days, we put the key strokes in double quotes. However, this convention was changed to AsciiDoc's `kbd:` syntax later. Apparently, we missed some key strokes during the migration. So, this commit changes the remaining ones to the proper syntax.

Lowercase keys in the old syntax are now converted to uppercase. The old uppercase keys still remain uppercase, but get an explicit Shift prefix.


Minoru (approved):
Thank you!

Issue #2989, opened 2025-02-09
Assertion hit when setting unknown variable (Assertion `0 && "unreachable"' failed)


dennisschagt:
**Newsboat version**: c39f99cbb8ba785f071b93e3bb0ec886cfaacf27 (current master)

Steps to reproduce the issue:
1. Open Newsboat's cmdline (press colon `:`)
2. Set a non-existent variable (`set feedlist-title abc`)
3. Press enter and observe that Newsboat crashes with `Assertion `0 && "unreachable"' failed.`

**Other info you think is relevant**:
Full error:
```
newsboat: src/configdata.cpp:85: nonstd::expected_lite::expected<void, std::__cxx11::basic_string<char> > newsboat::ConfigData::set_value(std::string): Assertion `0 && "unreachable"' failed.
```

PR #2990, opened 2025-02-09
Avoid crash on setting non-existent config option

dennisschagt:
Resolves https://github.com/newsboat/newsboat/issues/2989


dennisschagt:
For reference, example output:
image

PR #2991, opened 2025-02-09
Fix handling of operations without default key

dennisschagt:
The operations with an empty default key were still being registered.
That has no user visible effect (as it is not possible to input an empty keypress).
However, it does interfere with further changes I want to make so fixing this first.

### DIFF
--- a/src/keymap.cpp
+++ b/src/keymap.cpp
@@ -49,7 +49,7 @@ namespace newsboat {
 struct OpDesc {
 	const Operation op;
 	const std::string opstr;
-	const KeyCombination default_key;
+	const nonstd::optional<KeyCombination> default_key;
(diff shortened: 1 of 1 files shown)


src/keymap.cpp
```diff
@@ -743,7 +743,7 @@ static const std::vector<OpDesc> opdescs = {
 	{
 		OP_INT_GOTO_URL,
 		"gotourl",
-		KeyCombination("internal-goto-url"),
+		{},
```
dennisschagt, line 746: Also replace `KeyCombination("internal-set")` above (can't comment on it directly due to github comment restriction)

dennisschagt: Done

PR #2996, opened 2025-02-15
Reduce raw pointer usage

dennisschagt:
I started experimenting a bit with jj (Jujutsu).
That made me aware of some old commits/branches I had floating around.

Going through those now to see if there is still anything useful.


dennisschagt:
I don't see any risks with these simple refactors so I will just enable auto-merge directly.

PR #2997, opened 2025-02-16
Keybindings: Allow escaping `<` and `>` using `<LT>` and `<GT>`

dennisschagt:
Part of https://github.com/newsboat/newsboat/pull/2364

### DIFF
--- a/rust/libnewsboat/src/keycombination.rs
+++ b/rust/libnewsboat/src/keycombination.rs
@@ -83,7 +83,12 @@ fn alphabetic(input: &str) -> IResult<&str, char> {
 }
 
 fn regular_bindkey(input: &str) -> IResult<&str, KeyCombination> {
-    Ok(("", KeyCombination::new(input.to_owned())))
+    let key_combination = match input {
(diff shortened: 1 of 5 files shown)


Minoru (commented):
I must be missing some context, can you please clarify? From the title I assumed one would be able to write `bind 2<LT>4 open` and type <kbd>2</kbd><kbd>&lt;</kbd><kbd>4</kbd> to invoke `open`. However, the implementation seem to be about the internal representation of the `<` and `>` keys. Did I misunderstand the purpose of the PR?

(This is excellent work from the technical point of view, I just don't fully understand what's being accomplished.)

dennisschagt:
> From the title I assumed one would be able to write bind 2<LT>4 open and type 2<4 to invoke open.

That is indeed possible now.
The changes indeed focus on how bindings with `<` and `>` are stored.
However, by storing it that way, uses with the fully written out variant (`<LT>`/`<GT>`) will end up in that storage in the same way.
So it is mainly work on normalization of `<` and `<LT>` to store them in the same way.

PR #3002, opened 2025-02-18
Highlight search phrase case insensitively

### LINKED ISSUE #2998: Highlighting in HelpFormAction is case sensitive
**Newsboat version**: Current master (9189eb858305672fbbf6aa96193d2520853a6a13)

Steps to reproduce the issue:

1. Open help window (press `?`)

2. Search for "open" (`/open<ENTER>`)

3. Notice that lines containing "open" and "Open" show up (as expected) but only the "open" occurrences with the same capitalization are highlighted

Image

juntuu:
I think case-insensitive regex search & replace is suitable for this.

There could still be some discrepancy between the case matching of this and `strcasestr`, which is used as the precondition check.

The added function is quite specific. It might also be beneficial to construct the regex once outside the `apply_highlights` lambda, but don't know if it's worth the bridging effort.

Fixes #2998

### DIFF
--- a/src/helpformaction.cpp
+++ b/src/helpformaction.cpp
@@ -103,12 +103,12 @@ void HelpFormAction::prepare()
 				|| strcasestr(desc.desc.c_str(), searchphrase.c_str()) != nullptr;
 		};
 
-		std::string highlighted_searchphrase = strprintf::fmt("<hl>%s</>", searchphrase);
 		const auto apply_highlights = [&](const std::string& line) {
+			auto text = StflRichText::from_plaintext(line);
 			if (apply_search && searchphrase.length() > 0) {
-				return utils::replace_all(line, searchphrase, highlighted_searchphrase);
+				text.highlight_searchphrase(searchphrase);
 			}
-			return line;
+			return text;
 		};
 
 		for (const auto& desc : bound_descriptions) {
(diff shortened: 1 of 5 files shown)


juntuu:
Seems like there are more issues with stfl quoting. Searching for `<` or `>` misbehaves, as does highlighting next to those characters.

Not sure if broken in these changes, or somewhere else. I'll investigate later, setting as a draft for now.

juntuu:
I changed the approach a bit, and made the highlighting a method on StflRichText. To me this feels nicer for the use case.

src/helpformaction.cpp
```diff
@@ -166,11 +160,11 @@ void HelpFormAction::prepare()
...
 				const std::string description = macro.second.description;
 
-				// "macro-prefix" is not translated because it refers to an operation name
-				std::string line = strprintf::fmt("<macro-prefix>%s  %s", key, description);
-				line = utils::quote_for_stfl(line);
-				line = apply_highlights(line);
-				listfmt.add_line(StflRichText::from_quoted(line));
+				if (should_be_visible({ macro.first, "", description, "", 0 })) {
+					// "macro-prefix" is not translated because it refers to an operation name
+					std::string line = strprintf::fmt("<macro-prefix>%s  %s", key, description);
```
juntuu, line 165: There is still a minor mismatch here. The `<macro-prefix>` is not part of the search, but it is part of the highlighting.

I suppose it is fine, though.

dennisschagt (approved):
Nice, LGTM

Issue #2492, opened 2023-07-12
Add feed-sort-order like lastupdated, but only consider New/unread [feature]


quite:

quite:
Hi @dennisschagt @Minoru I filed this a while back. Now made an attempt at implementing it.

quite:
I'm making some comments in the PR.

PR #3010, opened 2025-02-24
Implement latestunread sorting

quite:
Fixes https://github.com/newsboat/newsboat/issues/2492


src/feedcontainer.cpp
```diff
@@ -109,6 +109,45 @@ void FeedContainer::sort_feeds(const FeedSortStrategy& sort_strategy)
...
+				return result;
+			}
+			std::vector<std::shared_ptr<RssItem>> a_unread;
+			std::copy_if(a->items().begin(), a->items().end(),
+				std::back_inserter(a_unread),
+			[](const std::shared_ptr<RssItem>& item) {
+				return item->unread();
+			});
+			std::vector<std::shared_ptr<RssItem>> b_unread;
+			std::copy_if(b->items().begin(), b->items().end(),
```
quite, line 130: So this makes a filtered copy. I'm thinking that this is kind of inefficient, but also that efficiency does not matter so much here. From what I understand, newsboat is quite lazy in re-sorting. For example, there is no re-sorting when number of unread changes (considering unreadarticlescount sorting). Aside from restarting newsboat, it seems re-sorting can be provoked by changing which sort method is used (`g` key).

dennisschagt: From my own local testing, it indeed does not feel slow at all.
However, my list of feeds, and especially my list of unread feeds, is relatively small.
In general, I prefer to write things in an efficient way, but I would be fine with merging this as-is.

I did some experimenting locally with an alternative approach.
I might share it a bit later, still want to test it a bit more extensively.

src/feedcontainer.cpp
```diff
@@ -109,6 +109,45 @@ void FeedContainer::sort_feeds(const FeedSortStrategy& sort_strategy)
 			}
 		});
 		break;
+	case FeedSortMethod::LATEST_UNREAD:
+		std::stable_sort(
+			feeds.begin(), feeds.end(),
+		[&](std::shared_ptr<RssFeed> a, std::shared_ptr<RssFeed> b) {
+			if (a->unread_item_count() == 0 || b->unread_item_count() == 0) {
```
dennisschagt, line 116: If I understand correctly, the current implementation sorts feeds by their most recent unread articles.
For feeds which have no unread articles, the order is kept unchanged.
(except in case of reverse sort, where it seems the order is flipped every time it is selected)

I'm wondering if it makes sense to sort the "read" feeds as well based on their most recent article.
I've experimented with that for a bit locally, I might share that in a bit.

src/feedlistformaction.cpp
```diff
@@ -127,11 +127,11 @@ bool FeedListFormAction::process_operation(Operation op,
 		// i18n: This string is related to the letters in parentheses in the
 		// "Sort by (f)irsttag/..." and "Reverse Sort by
 		// (f)irsttag/..." messages
-		std::string input_options = _("ftauln");
+		std::string input_options = _("ftaulsn");
```
dennisschagt, line 130: I've been a bit too slow with reviewing.
We are currently preparing for a release.
One of the steps we take is freezing the translatable strings so I will wait with merging this PR until after the release (planned somewhere around March 23rd)

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Lists below only mention user-visible changes, but the full list of contributors
 for this release also includes TK
 
 ### Added
+- Ability to sort feedlist by latest unread (#2492) (Daniel Lublin)
```
dennisschagt, line 9: In general, we leave the changelog writing until release.
@Minoru Would it help you if PRs would include an update to the changelog or do you prefer writing them as part of the release process?

dennisschagt (approved):
I will postpone merging this until after the newsboat 2.39 release, but overall looks good to me

src/feedcontainer.cpp
quite: Yes, I think also sorting feeds with no unread on latest "read" can make sense.

quite:
Just rebased this

CHANGELOG.md
quite: I removed any changelog update by this pr

src/feedcontainer.cpp
dennisschagt: I haven't really found time to test my suggested changes, so I will instead merge this as-is (and maybe apply the improvement at a later time)

dennisschagt (approved):
Looks good, thank you for your patience!

PR #3012, opened 2025-03-01
HelpForm: Show new bindings

dennisschagt:
Part of https://github.com/newsboat/newsboat/pull/2364

Example output:
image

### DIFF
--- a/include/keymap.h
+++ b/include/keymap.h
@@ -2,9 +2,11 @@
 #define NEWSBOAT_KEYMAP_H_
 
 #include <map>
+#include <set>
 #include <string>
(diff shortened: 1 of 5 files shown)


dennisschagt:
Seeing it in practice, I think it might be nicer to show `<` instead of `<LT>` in the help view.
Especially given that it is unlikely to cause confusion (there is very little reason for users to use `<` and `>` in such a way in which it causes confusion)

der-lyse:
Thanks really cool, thanks!

Agreed, showing literally `<` and `>` instead of their encoded forms would look nicer.

PR #3022, opened 2025-03-09
chore(l10n): Update Polish translation before 2.39 release

carno:
Ref: https://github.com/newsboat/newsboat/pull/3020#issuecomment-2708990664


der-lyse (approved):
Thanks!

PR #3027, opened 2025-03-12
Update Dutch translation

dennisschagt:


po/nl.po
```diff
@@ -976,9 +976,9 @@ msgid "usage: set <config-option> <value>"
 msgstr "gebruik: set <configuratie-optie> <waarde>"
 
 #: src/formaction.cpp:252
-#, fuzzy, c-format
+#, c-format
 msgid "error setting '%s' to '%s': %s"
-msgstr "Fout tijdens het zoeken naar ‘%s’: %s"
+msgstr "Fout tijdens instellen van '%s' op ‘%s’: %s"
```
der-lyse, line 981: The first quotes seem to be ASCII single quotes, but the second are typographic ones. I'd naively expected them to both use the same, but maybe this has some internal logic in the Dutch translations.

der-lyse (approved):
Thanks!

po/nl.po
dennisschagt: That was not intentional, no idea how I got those in there

PR #3082, opened 2025-06-06
Port generate binary to Rust

dennisschagt:
There are more of these binaries we can port but I first want to check for any issues with this first one (either from review or from CI).
It might also be nice to use a more descriptive name than just "generate". I haven't come up with a good name yet, so suggestions are welcome.


dennisschagt:
> Speaking of awk scripts, I think awk might be more suitable for porting these c++ scripts to:

Nice idea, suggestion applied via https://github.com/newsboat/newsboat/pull/3102

PR #3092, opened 2025-06-11
Update German translations

der-lyse:
I took inspiration from the [Dutch translations](https://github.com/newsboat/newsboat/pull/3091/files#diff-44a066469e7c75b3cdbbe18743cae1045124dcb86d20c65269ba4c323b04f61cR749) on the properly space-separated sort options instead of jamming it together like in the English source. That's much nicer, especially with longer words in German. Even though the German noun "Titel" ideally should have been title-cased ;-) as well, I left it unchanged to be backwards-compatible and not break a user's muscle memory.

I couldn't verify the attempted fix in the second commit right now, but I'm pretty sure it should work. I only noticed that when looking at the code. (I never use that sorting feature myself.)


dennisschagt (approved):
Thanks, translation looks good and nice that you caught the missing check updates!

PR #3093, opened 2025-06-14
Update ES translation

Roboron3042:


po/es.po
```diff
@@ -728,27 +728,25 @@ msgstr "¡No hay fuentes seleccionadas!"
...
 #: src/feedlistformaction.cpp:132
-#, fuzzy
 msgid ""
 "Sort by (f)irsttag/(t)itle/(a)rticlecount/(u)nreadarticlecount/(l)astupdated/"
 "late(s)tunread/(n)one?"
 msgstr ""
-"¿Ordenar por (p)rimera etiqueta/(t)ítulo/número de (a)rtículos/número de "
-"artículos (s)in leer/ú(l)timo actualizado/(n)ada?"
+"¿Ordenar por (p)rimera etiqueta/(t)ítulo/(n)úmero de artículos/número de "
+"artículos (s)in leer/último (a)ctualizado/último (l)eído/(n)ada?"
```
der-lyse, line 740: You can't have the same key assigned twice. This will always cause the "article count" to be selected when pressing <kbd>n</kbd>, but it's impossible to actually choose "none".

Another thing to keep in mind, changing the keys might break the user's muscle memory. I'm not saying you should never do it, but it's probably worth thinking twice about it before committing to the reassignment. :-)

der-lyse (changes requested):
Thanks! You just have to think of a different key. Other than that, this looks good to me.

Issue #3115, opened 2025-07-13
OPML import destroys query feeds in urls file


der-lyse:
**Newsboat version (copy the output of `newsboat -v` or the first line of `git show`)**:

```
Newsboat r2.40-22-g9bba
System: Linux 5.10.0-32-amd64 (x86_64)
Compiler: g++ 10.2.1 20210110
ncurses: ncurses 6.2.20201114 (compiled with 6.2)
libcurl: libcurl/7.74.0 OpenSSL/1.1.1w zlib/1.2.11 brotli/1.0.9 libidn2/2.3.0 libpsl/0.21.0 (+libidn2/2.3.0) libssh2/1.9.0 nghttp2/1.43.0 librtmp/2.3 (compiled with 7.74.0)
SQLite: 3.34.1 (compiled with 3.34.1)
libxml2: compiled with 2.9.10
```

While working on #3112, I noticed a urls file corruption when importing feeds from OPML. Steps to reproduce the issue:

1. Start with an urls file like:

```
"query:New Articles:unread = \"yes\""
"query:All Articles:total_count > 0"

# Blogs
https://blog.fefe.de/rss.xml?html news blog security
```

2. Import an OPML file, e.g. one from our test data:

```
$ LC_ALL=C ./newsboat -i test/data/example.opml 
Import of test/data/example.opml finished.
```

3. Start Newsboat again, it aborts with a complaint about a broken urls file:

```
$ LC_ALL=C ./newsboat                                       
Starting Newsboat r2.40-22-g9bba...
Loading configuration...done.
Opening cache...done.
Loading URLs from /home/lyse/.newsboat/urls...done.
Loading articles from cache...Error while loading feed 'query:New': too few arguments
```

The urls file now looks like that:

```
query:New Articles:unread = "yes"
query:All Articles:total_count > 0 
https://blog.fefe.de/rss.xml?html "news" "blog" "security"
https://example.com/feed.xml
https://example.com/mirrors/distrowatch.rss
https://example.com/feed.atom
https://example.com/feed.rss09
https://blogs.example.com/~john/posts.rss "Blogs"
https://blogs.example.com/~mike/.rss "Blogs/friends"
https://fred.example.com/writing/index.php?type=rss "eloquent"
```

The double quotes around the query feeds are completely gone. But they are required, because the query feed name contains spaces. [According to our docs](https://newsboat.org/releases/2.40/docs/newsboat.html#_query_feeds) they're also needed when the filter expression contains spaces.

[...]

PR #3116, opened 2025-07-13
Document quotes when query feed name contains spaces

der-lyse:
Discovered via #3115.


dennisschagt:
Thanks, nice improvement

PR #3162, opened 2025-09-07
Update Dutch translation

dennisschagt:


dennisschagt:
Not sure. The coveralls comments are still coming in so it is not completely broken.
We also still have the `Coveralls / Calculate test coverage (pull_request)` check completing successfully.
@Minoru Is there any specific reason to keep the `coveralls/coveralls` check marked as required? (given that we already have the other Coveralls check marked as "required")

Issue #3204, opened 2025-10-26
Position in article only updated on window resize [regression]


dorsiflexion:
### Newsboat version (copy the output of `newsboat -v` or the first line of `git show`):

```
Newsboat r2.41
System: Linux 6.15.11-hardened1-1-hardened (x86_64)
Compiler: g++ 15.2.1 20250813
ncurses: ncurses 6.5.20240427 (compiled with 6.5)
libcurl: libcurl/8.16.0 OpenSSL/3.6.0 zlib/1.3.1 brotli/1.1.0 zstd/1.5.7 libidn2/2.3.7 libpsl/0.21.5 libssh2/1.11.1 nghttp2/1.67.1 nghttp3/1.12.0 (compiled with 8.16.0)
SQLite: 3.50.4 (compiled with 3.50.4)
libxml2: compiled with 2.15.0
```

Terminal emulator: alacritty 0.15.0-dev

### Config file (copy from ~/.newsboat/config or ~/.config/newsboat/config):

I don't think anything special is necessary. Here's my minimal example though:
```
urls-source "ocnews"
ocnews-url "http://****"
ocnews-login "*****"
ocnews-password "*******"
ssl-verifypeer no
ssl-verifyhost no
```

### Steps to reproduce the issue:

1. Open an article.

2. Scroll down.

3. Position in hints bar keeps displaying `Top` instead of `10%`, `20%` etc.

The article position should be updated when I scroll. But it doesn't. It does update when I resize the window though.

### Other info you think is relevant:

This is a regression because the position indicator worked previously.

dorsiflexion:
Git bisect spits out 4747a57a as the culprit:

```
# first bad commit: [4747a57af76b992c15b3ac4a06549f134399f1b9] Show error when current formaction doesn't support requested operation
```

PR #3205, opened 2025-10-26
Fix missing position update in item view

juntuu:
fixes: #3204


dennisschagt (approved):
Nice, thank you for the quick fix!

PR #3224, opened 2025-12-03
Add support for saving in Miniflux

wzykubek:
Saving is like bookmarking, but the article is sent to a configured read-it-later service instead.

Miniflux has support for many different external services, like Wallabag, Linkwarden, Shiori, etc., for saving articles to read them later. This PR adds support for this feature in the same way that current `miniflux-flag-star` works. I really missed this feature in Newsboat, so I have added it.


dennisschagt (approved):
Thanks! Looks good

dennisschagt:
It looks like we ran out of free Cirrus credits at the end of last month.
Additionally, we ran into a small issue with Curl which got resolved on the master branch already.
I've rebased your PR to take in the fix/workaround and trigger a new CI run.

PR #3256, opened 2026-01-18
Reduce duplication in Regexmanager

dennisschagt:


dennisschagt:
I'm thinking about implementing the "inherit" idea from https://github.com/newsboat/newsboat/issues/2054.
This PR and one or more followup PRs are purely refactoring/cleanup in preparation.
I intend to merge these relatively quickly as test coverage is pretty good for these parts and they don't change behavior.
If/when I create a PR for the "inherit" idea, I will leave that open for review for a while (specifically because it would be a behavior change)

Issue #3266, opened 2026-02-08
Possible memory leakage in `Regex::new` and `Regex::matches`


rikkaii:
Hi, I found two memory issue in this crate: Calling `pattern.into_raw()` will discard ownership of the memory in the `CString` variable `pattern`, which may lead to a memory leak. I think this can use `pattern.as_ptr()` instead. Please see the document for more detail  [CStrng::into_raw](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.into_raw)

https://github.com/newsboat/newsboat/blob/b067d136c4d1984bbc476d536522ef992ea0408a/rust/regex-rs/src/lib.rs#L154

The following code also has same problem:

https://github.com/newsboat/newsboat/blob/b067d136c4d1984bbc476d536522ef992ea0408a/rust/regex-rs/src/lib.rs#L204

dennisschagt:
Looks like you're right, thanks for the report!

There are indeed 2 uses of memory leaking `into_raw()` in our regex implementation and it looks like we have 3 more uses in other components (but those 3 are part of code which is currently unreachable: related to `MatcherError` FFI)

It looks like this leak can also be detected by using AddressSanitizer with Rust:
```
LSAN_OPTIONS='report_objects=1' RUSTFLAGS='-Zsanitizer=address' RUSTDOCFLAGS='-Zsanitizer=address' cargo +nightly test -p regex-rs --lib
```
(we already run ASAN with C++ but apparently not yet with Rust)

Out of curiosity: How did you find the issue?

PR #3271, opened 2026-02-15
Avoid `CString::into_raw()` uses which leak memory

dennisschagt:
Fixes https://github.com/newsboat/newsboat/issues/3266

There are 3 more uses of `into_raw()` in `matcher_error_to_ffi()`.
It looks like we cannot simply replace those by `as_ptr()` because the data is actually returned to our C++ code which requires a longer lifetime for the `CString`.
I will leave it as-is for now. I rather spend time to convert that part of the code to use `cxx` for FFI instead.
For now, I believe that part of the code is not reachable (not yet used from the C++ side)

---

We might be able to detect these kinds of memory leaks automatically by using AddressSanitizer for Rust as well, e.g.:
```
LSAN_OPTIONS='report_objects=1' RUSTFLAGS='-Zsanitizer=address' RUSTDOCFLAGS='-Zsanitizer=address' cargo +nightly test -p regex-rs --lib
```

I ran into a few issues while playing around with AddressSanitizer on our Rust code:
- On my machine I'm getting a nice stack trace for every memory leak. In the ubuntu 24.04 docker image (instantiated with our CI ASAN config) I'm not seeing any stacktraces.
- On my machine, gettext-rs fails to build with the above ASAN run. I haven't looked into the exact issue yet. For now, I've run the ASAN checks with gettext references locally removed from regex-rs.
- When trying to run ASAN checks on libnewsboat, I'm running into compilation issues in the `zerofrom` crate. This only happens if I add the `RUSTFLAGS` envvar. Without that envvar, libnewsboat compiles fine on Rust nightly.
  I will see if I can reproduce this outside of the Newsboat repositoy and report an issue on the Cargo repo if necessary.

### DIFF
--- a/rust/regex-rs/src/lib.rs
+++ b/rust/regex-rs/src/lib.rs
@@ -151,7 +151,7 @@ impl Regex {
 
         unsafe {
             let mut regex: regex_t = mem::zeroed();
-            let errcode = regcomp(&mut regex, pattern.into_raw(), flags.bits());
+            let errcode = regcomp(&mut regex, pattern.as_ptr(), flags.bits());
 
             if errcode == 0 {
                 Ok(Regex { regex })
@@ -201,7 +201,7 @@ impl Regex {
 
             regexec(
                 &self.regex,
-                input.into_raw(),
+                input.as_ptr(),
                 max_matches as libc::size_t,
                 pmatch.as_mut_ptr(),
                 flags.bits(),


dennisschagt:
I'm running into this error when trying to merge this PR:
image

Somehow Github does not show the missing required check in the build status.
I will see if a rebase onto https://github.com/newsboat/newsboat/pull/3273 will resolve this.

PR #3283, opened 2026-02-28
CI: add FreeBSD 15, replace Clang 16/17/18 with 19/20/21

Minoru:


Minoru:
I requested a review because the changes to FreeBSD aren't the simple `s/old version/new version/`.

PR #3286, opened 2026-03-01
Add job for running AddressSanitizer on Rust code

### LINKED ISSUE #3278: Run AddressSanitizer for Rust libraries in CI
**Newsboat version**: Current master (c532779452f7d67e0603e519f9de2d59d123aecc)

We recently had a report for a memory leak related to our use of `CString::into_raw()`.
I believe we can catch these kinds of issues automatically in CI by running our tests with AddressSanitizer (or just leak sanitizer) enabled for the Rust libraries.

I've described a few findings from running these checks locally.
See top post and @Minoru's first reply: https://github.com/newsboat/newsboat/pull/3271

dennisschagt:
Resolves https://github.com/newsboat/newsboat/issues/3278

It looks like Ubuntu's `llvm` package makes llvm-symbolizer available as binary with name `llvm-symbolizer-18`.
Adding `llvm-symbolizer` to the path based on [Rust sanitizers docs](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html#symbolizing-the-reports):
> Sanitizers produce symbolized stacktraces when llvm-symbolizer binary is in PATH.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -267,7 +267,7 @@ task:
     before_cache_script: *before_cache_script
 
 task:
-    name: AddressSanitizer (Clang 21, Ubuntu 25.10)
+    name: AddressSanitizer C++ (Clang 21, Ubuntu 25.10)
     skip: "changesIncludeOnly('contrib/*', 'contrib/**/*', 'doc/*', 'doc/**/*', 'docker/remote-testing/*', 'docker/remote-testing/**/*', 'po/*', 'po/**/*', 'snap/*', 'snap/**/*')"
 
     container:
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -262,6 +262,35 @@ task:
...
+    env:
+      RUSTFLAGS: "-Zsanitizer=address"
+      RUSTDOCFLAGS: "-Zsanitizer=address"
+      LSAN_OPTIONS: "report_objects=1"
+      JOBS: 8
+
+    cargo_cache: *cargo_cache
+
+    # We don't pass --keep-going to the build: failures can be debugged with
+    # logs of other jobs.
```
Minoru: This comment was probably copied from another task and doesn't make sense here.

Minoru (approved):
Looks good!

Don't we need the same symlink hack for the C++ job?

Minoru:
Thank you! I'll make the new job required now.

PR #3295, opened 2026-03-11
Update Russian and Ukrainian PO files

Minoru:


Minoru:
Enabling auto-merge because this is essentially a typographic edit—just a few new quote marks.

PR #3319, opened 2026-04-06
Check if all attributes are valid before applying filter

### LINKED ISSUE #2857: List contains stale items after applied filter triggers a MatcherException

dennisschagt:
Resolves https://github.com/newsboat/newsboat/issues/2857
Obsoletes https://github.com/newsboat/newsboat/pull/2858

### DIFF
--- a/src/feedlistformaction.cpp
+++ b/src/feedlistformaction.cpp
@@ -1155,8 +1155,8 @@ void FeedListFormAction::apply_filter(const std::string& filtertext)
 			do_redraw = true;
 		} else {
 			v.get_statusline().show_error(strprintf::fmt(
-					_("Error: Some filter attributes not availabe in feedlist: %s"),
-					utils::join(unavailable_attributes, ",")));
+					_("Error: Some filter attributes are not availabe in feedlist: %s"),
+					utils::join(unavailable_attributes, ", ")));
 		}
 	}
 }
--- a/src/feedlistformaction.cpp
+++ b/src/feedlistformaction.cpp
@@ -1132,15 +1132,32 @@ void FeedListFormAction::apply_filter(const std::string& filtertext)
 	}
 
 	filterhistory.add_line(filtertext);
-	if (!matcher.parse(filtertext)) {
+
+	Matcher new_matcher;
+	if (!new_matcher.parse(filtertext)) {
 		v.get_statusline().show_error(strprintf::fmt(
 				_("Error: couldn't parse filter expression `%s': %s"),
 				filtertext,
 				matcher.get_parse_error()));
 	} else {
-		save_filterpos();
-		filter_active = true;
-		do_redraw = true;
+		const auto referenced_attributes = new_matcher.get_referenced_attributes();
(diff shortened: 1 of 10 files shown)


Minoru:
I'll take a proper look tomorrow, but one question I have already is how this impacts the Rust post of filter parsing. Would it need updating? IIRC it only does parsing and evaluation, but it has to interface with `Matcher` too. (I'm not asking for that Rust code to be updated, I just want to know if it would need updating later.)

src/feedlistformaction.cpp
```diff
@@ -1132,15 +1132,32 @@ void FeedListFormAction::apply_filter(const std::string& filtertext)
...
+
+		if (unavailable_attributes.empty()) {
+			matcher = new_matcher;
+			save_filterpos();
+			filter_active = true;
+			do_redraw = true;
+		} else {
+			v.get_statusline().show_error(strprintf::fmt(
+					_("Error: Some filter attributes not availabe in feedlist: %s"),
+					utils::join(unavailable_attributes, ",")));
```
dennisschagt: From Lyse on irc:
> Are the two new error messages missing an "are"?
> Also, not sure if joining the unavailable attributes with comma and space would make them more readable.

I think those are nice suggestions so I plan to take them along

dennisschagt (commented):
> one question I have already is how this impacts the Rust post of filter parsing. Would it need updating? IIRC it only does parsing and evaluation, but it has to interface with Matcher too. (I'm not asking for that Rust code to be updated, I just want to know if it would need updating later.)

I think it does, but looks like it is quite straightforward to port this.
Untested, but I expect something like this would suffice:
```
impl Matcher {
    // ...

    pub fn get_referenced_attributes(&self) -> HashSet<&str> {
        let mut attributes: HashSet<&str> = HashSet::new();
        let mut to_visit = vec![&self.expr];
        while let Some(expr) = to_visit.pop() {
            match expr {
                And(l, r) => {
                    to_visit.push(l);
                    to_visit.push(r);
                }
                Or(l, r) => {
                    to_visit.push(l);
                    to_visit.push(r);
                }
                Comparison { attribute, .. } => {
                    attributes.insert(attribute);
                }
            }
        }
        attributes
    }

}

```

Minoru (approved):
Looks great!

PR #3322, opened 2026-04-13
Fix typo in spelling of "available"

DannyJJK:
I noticed in a recent commit (e20b308) there was a spelling mistake on the word "available". I've fixed this, and also found another area where the same typo was made in a comment which I've fixed as well.


Minoru (approved):
Thanks!

PR #3344, opened 2026-05-24
Migrate from Cirrus CI to GitHub Actions

### LINKED ISSUE #3326: Migrate away from Cirrus CI
Cirrus CI, who served us great over the years, is shutting down: https://cirruslabs.org/ ([Web Archive](https://web.archive.org/web/20260412063228/https://cirruslabs.org/))

TODO:

- [ ] look for CI providers that offer FreeBSD for free for open source

Minoru:
As the title says, this replaces Cirrus CI with GitHub Actions.

### The approach

Cirrus CI had this feature called "Dockerfile-as-environment" where we specify a Dockerfile and Cirrus takes care of building an image for us, then runs our commands in a container with that image. GitHub Actions doesn't have a matching feature, but it provides various building blocks.

I guess the "idiomatic" approach would be to lean on GitHub Container Registry: let our CI build an image, push it into our registry, then another job uses those images as environment. I shied away from this approach because I didn't want to be bothered with learning how GHCR access works, who can push, who can pull, how we give CI those rights etc.

The approach taken in this PR follows what I'd do locally: build the image with `docker build`, then run my commands with `docker run`. It looks quite ugly, and `docker build` is replaced by `docker buildx` wrapped in a third-party Action so I don't have to do the caching manually.

GHCR would remove the ugliness of `docker run`, so I might still look into it, but not right now.

### Things that don't work

1. ~/.cargo caching on FreeBSD. Those jobs start a VM inside GitHub Actions runner, copy the code in there, and run the build commands via SSH. In order to cache ~/.cargo, we'd need to copy files back to the runner, and the action even does that by default, but it was failing for me and I disabled it. Something to re-visit later: https://github.com/newsboat/newsboat/issues/3366
2. Docker image caching for "MSRV GCC 14" job. It's possible that it just didn't have a chance to set the cache yet because I changed the cache configuration a lot of times. I'll see how cache performs when CI is triggered for this PR.

### `Assisted-By`

Many commits are marked by "Assisted-by" tag because I ran a local LLM to do this change across all workflows:

```diff
--- a/.github/workflows/asciidoctor.yml
+++ b/.github/workflows/asciidoctor.yml
@@ -40,13 +40,13 @@ jobs:
             UID=${{ env.UID }}
             GID=${{ env.GID }}

-          cache-from: type=gha,scope=build=${{ matrix.name }}
-          cache-to: type=gha,scope=build=${{ matrix.name }},mode=max
+          cache-from: type=gha,scope=image
+          cache-to: type=gha,scope=image,mode=max

           push: false
           load: true

-          tags: local/container:latest
+          tags: newsboat/ci-20.04-asciidoctor:latest

       - name: Run Asciidoctor
         run: |
@@ -55,5 +55,5 @@ jobs:
             --tty \
             --mount type=bind,source=$(pwd),target=/home/builder/src \
             --env ASCIIDOCTOR_OPTS='--failure-level WARN' \
-            local/container:latest \
+            newsboat/ci-20.04-asciidoctor:latest \
             make doc
```

(It's buggy -- "image" in `scope=image` should be an image name e.g. `scope=newsboat/ci-20.04-asciidoctor`, but that's on me; the LLM did exactly what I instructed it to).

I didn't do it by hand because it's boring, and I didn't use `sed` because `sed` can't look at the job name and come up with a matching Docker image name. (I could probably write a program to do that, but it'd take way longer than doing everythin by hand, and it's not that interesting either, so I resorted to an LLM.)

All other changes are `Assisted-by: Vim 9.1` :)

### Timeline

Cirrus CI will stop working in one week (on June 1st), so the time to review this is limited. If problems are found that are too big for me to fix in time, I'll just disable jobs and work on fixes later.

Fixes #3326.


Minoru:
> Docker image caching for "MSRV GCC 14" job. It's possible that it just didn't have a chance to set the cache yet because I changed the cache configuration a lot of times. I'll see how cache performs when CI is triggered for this PR.

All the caches have been created anew. It's weird; according to https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache the CI job should have access to the caches of the branch.

I triggered a re-run of all checks that use Docker caching to verify that they *do* reuse caches.

Minoru:
Some jobs reused their caches and some did so only partially. Apparently we've exceeded the 10 GB limit of GitHub Actions caches; we're currently in some sort of grace period, using almost 14 GB. I guess we should optimise our Dockerfiles to better utilise the layering. This is a separate task really, so I'll file an issue instead of trying to address it here: https://github.com/newsboat/newsboat/issues/3365

.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,231 @@
...
+      - contrib/*
+      - contrib/**/*
+      - doc/*
+      - doc/**/*
+      - docker/remote-testing/*
+      - docker/remote-testing/**/*
+      - po/*
+      - po/**/*
+      - snap/*
+      - snap/**/*
```
dennisschagt: Aren't these equivalent to something like:

```suggestion
      - contrib/**
      - doc/**
      - docker/remote-testing/**
      - po/**
      - snap/**
```

Based on the [patterns-to-match-file-paths section](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#patterns-to-match-file-paths) of Github Action docs, especially these example matches:
Image

.github/workflows/clang_tidy.yml
```diff
@@ -0,0 +1,83 @@
...
+      - po/*
+      - po/**/*
+      - snap/*
+      - snap/**/*
+  pull_request:
+    paths_ignore: *ignored_paths
+
+jobs:
+  clang_tidy:
+    name: Run clang-tidy
```
dennisschagt, line 15: I expect the list of required jobs needs to be updated when this PR is merged:
Image

.github/workflows/freebsd.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Build, test, fake-install, run Clippy
+        uses: vmactions/freebsd-vm@v1
+        with:
+          release: ${{ matrix.release }}
+          copyback: false
+          prepare: |
+            # Update everything to avoid `pkg` looping on self-update check.
+            pkg update -f
+            pkg upgrade -y
```
dennisschagt, line 36: Is this still necessary here? (given there is a copy of these lines at line 78-79)

.github/workflows/linux.yml
```diff
@@ -0,0 +1,225 @@
...
+              cc=gcc-13
+              cxx=g++-13
+          - name: Rust stable, GCC 14 (Ubuntu 24.04)
+            image: newsboat/ci-24.04-gcc14:latest
+            dockerfile: docker/ubuntu_24.04-build-tools.dockerfile
+            build_args: |
+              cxx_package=g++-14
+              cc=gcc-14
+              cxx=g++-14
+          - name: Rust stable, GCC 15 (Ubuntu 25.04)
```
dennisschagt, line 75: Is there a reason to run this if we already run the same image but with more warnings and checks?

.github/workflows/linux.yml
```diff
@@ -0,0 +1,225 @@
...
+            image: newsboat/ci-25.04-gcc15:latest
```
dennisschagt, line 76: It looks like this image name is duplicated for the "more warnings and checks" instance.
So, both of those would use the same cache-to/cache-from value, based on:
```
          cache-from: type=gha,scope=${{ matrix.image }}
          cache-to: type=gha,scope=${{ matrix.image }},mode=max
```

I guess that is mostly fine because the docker build of those two matrix instances should produce the same image data to be stored in the cache.
Do you see any potential danger in the two matrix entries racing each other in creation/writing/loading from cache?

dennisschagt: Same duplication also holds for image name `newsboat/ci-25.10-clang21:latest` (2x in this matrix) and cache entry `newsboat/ci-25.10-clang21` (2x in sanitizers.yml)

dennisschagt (approved):
Nice work, thank you for picking this up before Cirrus becomes unavailable!

.github/workflows/sanitizers.yml
Minoru: I copy-pasted from Cirrus's config and it didn't occur to me that GHA might work differently in that regard. Will update; thanks!

.github/workflows/clang_tidy.yml
Minoru: That one's puzzling me because the name is actually the same as the name of the old job. GHA has this bug where changing the name of the workflow is only applied when it's in the master branch; I'm hoping that once the merge is done, it'll just line up automatically. If not, then I'll update the settings.

.github/workflows/freebsd.yml
Minoru: Yes. In between these lines and lines 78-79, we switch to a different repository, so the second set of commands might bring some updates.

You can see it in the logs:
* click "Build, test, fake-install, run Clippy"
* click "Run 'prepare' in VM"
* observe that there are two blocks that start with "Number of packages to be upgraded" -- that's two `pkg upgrade -y` commands running

.github/workflows/linux.yml
Minoru: No strong reason, more like a bunch of arguments to do it this way. In no particular order:

* this build is representative of how users and distributions would build Newsboat. "More warnings and checks" is less representative and is basically a testing ground for us. Keeping them separate makes it easy to say "we tested build from source on GCC 15 and it works", without adding the caveat "but we're building with some custom flags"
* the "more warnings and checks" job uses `-O3`. Higher optimisation levels increase the precision of warnings, which is to say, false positives might disappear and extra warnings might be triggered. In other words, this check might fail when the "more warnings..." one succeeds because a false positive is present at `-O2` but not at `-O3`. (This is theoretical so far, can't remember that ever happening in Newsboat, although I've seen it a few times in other code I worked on.)
* when a new GCC is out, we need to update the "more warnings..." job, add a new job for the ordinary build, and remove the oldest GCC job to keep their number in check. Currently it's easy to do; if we only had "more warnings and checks", there'd be a bit more shuffling of the config

Is there an argument to drop this job?

.github/workflows/linux.yml
Minoru: > Do you see any potential danger in the two matrix entries racing each other in creation/writing/loading from cache?

There's potential, but I haven't yet seen any problems in practice.

Docs for `docker buildx` and BuildKit don't seem to explain how caching works, but from reading the logs and looking at https://github.com/newsboat/newsboat/actions/caches I believe that it works like this:

* upon startup, `docker buildx` loads the "cache index" from the cache
* it reads layer IDs from the index and loads them from the cache too
* it builds the image, re-using layers if possible
* it uploads any new layers
* if there are new layers, it also uploads a new "cache index"

So the race on loading from cache is safe because everybody is getting the same thing. Creating/writing to the cache should also be safe as long as uploading is atomic (that's an assumption, I don't know if it really is).

Anecdotally, I haven't seen jobs failing because of loading a wrong cache or failing to upload one.

Re-using images is useful, so if you don't know of reasons why it won't work, I'd keep the image names duplicated for now.

.github/workflows/sanitizers.yml
Minoru: Done: https://github.com/newsboat/newsboat/pull/3344/commits/2d512cf163389a9782d4041349c51acfe469c164 I also removed `docker/` from ignored paths so CI rebuilds images and re-runs checks in them.

.github/workflows/linux.yml
dennisschagt: > Is there an argument to drop this job?

Not really.
I though it had no added value but your examples above make sense

dennisschagt (approved):
Nice, LGTM

.github/workflows/clang_tidy.yml
Minoru: > That one's puzzling me because the name is actually the same as the name of the old job

So this is not completely true: the name of the *workflow* is the same, but the *job* is just "Run clang-tidy". I think that's what tripped up GHA.

I now updated the required checks to match the new names.

PR #3347, opened 2026-05-24
Link to Colors section from relevant operations

### LINKED ISSUE #1873: Add informations about possible text attributes to documentation
# Version
**newsboat: 2.25.0**

# Problem
I was configuring my colorschemes and wanted to change text attributes. When consulting the man page all I got was:

```
color (parameters: <element> <fgcolor> <bgcolor> [<attribute> ...]; default value: n/a)
    Set the foreground color, background color and optional attributes for a certain element. (example: color background white black)
```
It says that I can apply an attribute, but nowhere in the documentation can be found list of valid attributes. So when I tried to change my text to *italics* all I got was message: `Error while processing command 'color listfocus color208 color234 italics': 'italics' is not a valid attribute`. I eventually found list of valid text attributes by cloning this project and then grepping for error messages and function names. This is far for ideal for a normal user.

# Solution
I believe that all it takes to fix this issue is to add to documentation (man page) section that would list possible attributes (and preferably short description to each entry).

dennisschagt:
Resolves https://github.com/newsboat/newsboat/issues/1873
Makes progress on https://github.com/newsboat/newsboat/issues/910

I also found the following attribute descriptions in [the ncurses manpages](https://man7.org/linux/man-pages/man3/curs_attr.3x.html) but I don't think the descriptions add much useful info over just listing the name:
image

Not all of the above attributes are available to us as STFL has a hardcoded string->attribute mapping (in [stfl/style.c](https://github.com/newsboat/stfl/blob/bbb2404580e845df2556560112c8aefa27494d66/style.c#L124-L146))


doc/newsboat.asciidoc
```diff
@@ -986,7 +986,7 @@ the text. If multiple highlight matches overlap, the style of the later
 specified highlight rule will be applied. `<fgcolor>` and `<bgcolor>` specify
 the foreground color resp. the background color of the matches. You can also
 specify 0 or more attributes. You can find a list of valid colors and
-attributes in the <<_configuring_colors>>.
```
Minoru, line 989: I'm surprised that "Asciidoctor with warnings-as-errors" job didn't catch this. Perhaps we can add some flag there to warn on dangling references?

This comment is not a blocker, feel free to merge if there isn't such a flag or it's too hard to add.

Minoru (approved):

doc/newsboat.asciidoc
dennisschagt: It looks like Asciidoctor does not have a robust way to check this automatically.
I've locally experimented a bit with the `-v` option (based on [Validate Cross References](https://docs.asciidoctor.org/asciidoc/latest/macros/xref-validate/)) but it gave a bunch of false positives.
It looks like patterns where an anchor is added and directly referenced on the same line cause a bunch of false positive (e.g. `[[articlelist-format-a]]<<articlelist-format-a,+a+>>`, which generates an INFO message under `-v` but works fine in practice)

A quick manual check did find a few broken references in our HTML output:
```bash
for anchor in $(cat doc/xhtml/newsboat.html | rg -o 'href="#([^"]*)"' -r '$1' | sort | uniq); do grep "id=\"$anchor\"" doc/xhtml/newsboat.html > /dev/null || echo $anchor; done
```
```
_configuring_colors
max-download
max-downloads
show-keymap-hints
```

dennisschagt:
Rebased to pull in the CI changes.
First commit stayed exactly the same, only added a new commit with a few broken link fixes:
image

PR #3368, opened 2026-05-31
Improve debug logs (by Lyse)

Minoru:
Patches submitted over IRC.

I already reviewed them and they look good, so I'm enabling auto-merge on this.

### DIFF
--- a/rss/parser.cpp
+++ b/rss/parser.cpp
@@ -136,7 +136,7 @@ nonstd::expected<Feed, Parser::Error> Parser::parse_url(const std::string& url,
 	if (etag_headers.size() >= 1) {
 		std::string etag = etag_headers.back();
 		utils::trim(etag);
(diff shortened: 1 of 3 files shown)


Minoru:
Weird. Three checks were (correctly) skipped because of their rules:

- .po files
- Asciidoctor with warnings-as-errors
- No dbg! macros in Rust code

Usually they'd be marked as skipped and won't prevent the PR from getting merged, but it's not what happened here. Instead it looks like GitHub is waiting on those checks even though we'll never run. I'm removing them from the required list for now and will look into this later.

Minoru:
GitHub docs describe the behaviour I'm expecting: 

> A job that is skipped will report its status as "Success". It will not prevent a pull request from merging, even if it is a required check.

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks

PR #3374, opened 2026-06-05
Update POT and PO files

Minoru:


Minoru:
Property tests found an issue with `LimitedString`, but it's unrelated to the changes in this PR because I can also reproduce the problem on master. Filed an issue about that (#3378) ; merging this because it's time to update the translations.

Minoru:
@newsboat/translators, please update your respective PO files by June 20th (two weeks from now). This time around we have 5 new messages and 1 fuzzy, so a bit more work than usual :) Thank you!

(In case you don't want to be pinged about Newsboat translations anymore, drop a comment here or email me.)

PR #3375, opened 2026-06-05
Update Russian and Ukrainian PO files

Minoru:


Minoru:
While reviewing other translation PRs I noticed people translate "argument" differently. Discussing that with translators I found it's a mistake. Now I checked my own translations and sure enough, the same mistake is there too :) The last two commits fix it. I'll probably squash the PR when merging.

Issue #1537, opened 2021-03-20
Reload queue file in podboat


tim-tx:
Would be great to have a command to reload the queue file in podboat and/or an autoreload timer. Use case is when running newsboat and podboat side by side, if you enqueue from newsboat, you have to exit and restart podboat to load the queue file.

dennisschagt:
I currently use the `pb-purge` command (default key: shift-p) for reloading.
It removes entries which are marked as finished/deleted but also reloads the queue file.

A separate `pb-reload` command would be nice though.

PR #3376, opened 2026-06-05
Implemented the `pb-reload` command (preserves items marked as finished/deleted)

hew02:
As was discussed in #1537, contributed a new podboat command `pb-reload` that essentially performs the same function of `pb-purge` but more clearly defines how to reload the Newsboat queue without restarting the program. Unlike purging, of course, reload simply just reloads the current state of the queue, not performing any cleanup work.


src/keymap.cpp
```diff
@@ -324,6 +324,13 @@ static const std::vector<OpDesc> opdescs = {
 		translatable("Mark download as deleted"),
 		KM_PODBOAT
 	},
+	{
+		OP_PB_RELOAD,
+		"pb-reload",
+		KeyCombination("r", ShiftState::Shift),
```
Minoru: Why did you pick this default? `^R` would be more in line with Newsboat; there it re-reads the urls file.

include/queueloader.h
```diff
@@ -25,7 +25,8 @@ class QueueLoader {
 	/// status are removed. If \a also_remove_finished is `true`, `FINISHED`
 	/// downloads are removed too.
 	void reload(std::vector<Download>& downloads,
-		bool also_remove_finished = false) const;
+		bool also_remove_finished = false,
+		bool also_remove_deleted = true) const;
```
Minoru, line 30: The doc needs to be updated.

Minoru (commented):
Thank you for the contribution! Can you write a test for this? We already got some in tests/queueloader.cpp, shouldn't be too hard to extend that.

src/keymap.cpp
hew02: Ahh I see that binding now, haven't used it before so that explains my ignorance!

hew02:
Thanks for taking the time to review! I've fixed those other points and writing the test now, still getting used to the intricacies of the project.

test/queueloader.cpp
```diff
@@ -104,6 +104,17 @@ TEST_CASE("reload() removes downloads iff they are marked as finished or deleted
 				}
 			}
 		}
+
+		WHEN("reload() is called with also_remove_finished == false and also_remove_deleted == false") {
```
Minoru, line 108: With two flags, there are four possible states. This test should check them all.

test/queueloader.cpp
```diff
@@ -464,6 +475,69 @@ TEST_CASE("reload() removes files corresponding to \"FINISHED\" downloads "
...
+	  "[QueueLoader]")
+{
+	test_helpers::TempFile queueFile;
+	auto empty_callback = []() {};
+
+	ConfigContainer cfg;
+
+	QueueLoader queue_loader(queueFile.get_path(), cfg, empty_callback);
+
+	test_helpers::TempFile fileInDeletedStatte;
```
Minoru:
```suggestion
	test_helpers::TempFile fileInDeletedState;
```

I see the same typo exists in other tests, can you fix it there as well?

Minoru (commented):
Almost there!

During my first review I missed the fact that the new operation is not documented. Please `git grep pb-purge` and update the file in doc/ where pb-purge is already documented.

hew02:
Ok I've done all those things, I guess I'm out of practice writing tests lol. Also what can I run besides astyle to check for formatting errors that the CI is tripping on? Running astyle on my system seems to create new egregious formatting errors.

Minoru:
Excellent, thank you!

> Also what can I run besides astyle to check for formatting errors that the CI is tripping on?

You can build the Docker image we use for formatting and run it locally: https://github.com/newsboat/newsboat/blob/dbf2d3eae9a7652673e8d89e8a2eec0d942dec87/docker/code-formatting-tools.dockerfile

Another option is to take a look at CI logs, copy the diff it prints out, and apply that.

hew02:
My bad, just saw it was set to auto-merge.

Minoru:
No problem; the merge wouldn't have happened anyway if formatting check didn't pass.

Minoru:
A-and merged! Congratulations on your first contribution to Newsboat, @hew02 :)

hew02:
Hopefully more to come!

PR #3380, opened 2026-06-07

carno:
Ref: https://github.com/newsboat/newsboat/pull/3374

### DIFF
--- a/po/pl.po
+++ b/po/pl.po
@@ -6,7 +6,7 @@ msgstr ""
 "Project-Id-Version: newsboat 2.5\n"
 "Report-Msgid-Bugs-To: https://github.com/newsboat/newsboat/issues\n"
 "POT-Creation-Date: 2026-06-05 19:02+0300\n"
-"PO-Revision-Date: 2025-12-12 21:26+0100\n"
+"PO-Revision-Date: 2026-06-08 23:08+0200\n"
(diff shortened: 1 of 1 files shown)


po/pl.po
```diff
@@ -968,19 +968,19 @@ msgid "couldn't parse filter expression `%s': %s"
...
+#, c-format
 msgid "Too many arguments for %s"
-msgstr "zbyt wiele parametrów."
+msgstr "Zbyt wiele parametrów dla %s"
 
 #: src/formaction.cpp:168
 #, c-format
 msgid "Too many arguments to \"cmdline\" command: %s"
-msgstr ""
+msgstr "Zbyt wiele argumentów dla polecenia: %s"
```
Minoru: I notice "arguments" are translated differently here. Is that correct?

Minoru: "cmdline" here is the name of the `cmdline` operation, so it's a bit suspicious that it's missing from the translation. Is that correct?

Minoru (commented):
Thanks! I spotted a few places that might be incorrect, although I don't know any Polish. Please take a look.

po/pl.po
carno: Simply sloppy on my side - both are correct but lets keep one.

carno: Fixed, Thanks!

Minoru:
Thank you!

PR #3382, opened 2026-06-09
Fix missing handling of tags by OpmlUrlReader

Jorenar:
Adds missing handling of `category` properties, and `text` from outer outlines

<sub>Probably worth taking a look at #292 sometime soon</sub>

### DIFF
--- a/src/opmlurlreader.cpp
+++ b/src/opmlurlreader.cpp
@@ -1,6 +1,7 @@
 #include "opmlurlreader.h"
 
 #include <cstring>
+#include <sstream>
 
 #include "logger.h"
 #include "utils.h"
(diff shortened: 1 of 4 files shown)


src/opmlurlreader.cpp
```diff
@@ -108,8 +122,10 @@ void OpmlUrlReader::rec_find_rss_outlines(xmlNode* node, std::string tag)
 				handle_node(node, tag);
 				xmlFree(type);
 			} else {
-				char* text = (char*)xmlGetProp(
-						node, (const xmlChar*)"title");
+				char* text = (char*)xmlGetProp(node, (const xmlChar*)"text");
+				if (!text) {
+					text = (char*)xmlGetProp(node, (const xmlChar*)"title");
+				}
```
Minoru, line 128: What's this? I haven't found `<text>` in any of the OPML specs.

Minoru (commented):
Can you please provide a test for categories? We already have some in test/opmlurlreader.cpp, shouldn't be too hard to create a new one.

src/opmlurlreader.cpp
Jorenar: Well, probably because it's `text=` not `<text>` ;-)

It mirrors what was made for `opml` class.
The property `text` is mandatory, while `title` is not.

Minoru: Ah.

Can we have a test for this too? Apparently we didn't add one in https://github.com/newsboat/newsboat/pull/3069 either, because `git grep 'text=' test/` only finds some unit tests for `opml::generate()`.

Jorenar: Sure, I'll change some `title=`s to `text=`s in test OPML files

Minoru:
Thank you!

PR #3383, opened 2026-06-09
Update Dutch translation

dennisschagt:


Minoru (approved):
Looks great; thanks!

GitHub Actions failed to reach Docker registry again :( It surprises me how bad it works. I restarted the jobs.

Minoru:
Actually, wait, why did Linux jobs run for this change at all? And where's the i18nspector job? I'll have to look into it on the weekend because it looks like I've broken something during migration to GHA.

PR #3385, opened 2026-06-14
Update of Swedish translation

deob83:
Translation of new and updated strings.

### DIFF
--- a/po/sv.po
+++ b/po/sv.po
@@ -6,7 +6,7 @@ msgstr ""
 "Project-Id-Version: newsboat-2.12\n"
 "Report-Msgid-Bugs-To: https://github.com/newsboat/newsboat/issues\n"
 "POT-Creation-Date: 2026-06-05 19:02+0300\n"
-"PO-Revision-Date: 2026-03-21 18:30+0100\n"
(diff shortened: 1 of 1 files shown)


po/sv.po
```diff
@@ -968,14 +969,14 @@ msgid "Operation \"%s\" not handled in dialog \"%s\""
...
+#, c-format
 msgid "Too many arguments for %s"
-msgstr "för många parametrar."
+msgstr "För många parametrar för %s"
 
 #: src/formaction.cpp:168
 #, c-format
 msgid "Too many arguments to \"cmdline\" command: %s"
-msgstr ""
+msgstr "För många argument för \"cmdline\"-kommando: %s"
```
Minoru, line 979: One of these translates "arguments" as "parametrar" while the other one translated it as "argument"; is that right?

deob83: Good catch! I'd say both works, but it's better to be consistent, so I've now changed the first one to "argument". I also found another instance where "arguments" had been translated with "parametrar", so I changed that as well.

Minoru:
Yeah, other translators (including myself) also ran into this. Thank you for the updates!

PR #3387, opened 2026-06-15
display an error if html-renderer failed to run

### LINKED ISSUE #280: Display an error if html-renderer failed to run

dud-eisme:
Closes #280 
Fix html-renderer error reporting via tuple propagation

Refactor item_renderer::to_stfl_list to return a std::tuple containing
the rendering status code along with the formatted text and line count.
This allows the UI layer in ItemViewFormAction to capture exit codes
(such as 127 for missing binaries or 42 for script errors) and cleanly
display localized error messages on the status line instead of falling
back to a generic execution failure.


3
Minoru, line 1: Looks like you committed a bunch of temporary files. Please remove them.

include/itemrenderer.h
```diff
@@ -45,7 +45,7 @@ std::string to_plain_text(
 /// highlight the resulting text. \a links is filled with all the links
 /// found in the article (which is useful for article view, which lets
 /// users open the links by their number).
-std::pair<std::string, size_t> to_stfl_list(
+std::tuple<std::string, size_t, int> to_stfl_list(
```
Minoru: A tuple is ill-suited for this because it's a combination of three fields, when in fact we want `string+size_t` OR an `int`. It's better to use `nonstd::expected<std::pair<std::string, size_t>, int>` here.

src/itemrenderer.cpp
```diff
@@ -174,6 +179,24 @@ void render_html(
...
+			} catch (...) {
+				exit_code = -1;
+			}
+		} else if (output.empty() && !source.empty()) {
+			exit_code = -1;
+		}
+
+		if (exit_code) {
+			return exit_code;
+		}
```
Minoru: This is a very round-about way of getting the exit code. Why not modify `utils::run_program` to return the exit code?

src/itemviewformaction.cpp
```diff
@@ -138,6 +140,23 @@ void ItemViewFormAction::prepare()
...
+				} else if (renderer_status == -1) {
+					v.get_statusline().show_error(strprintf::fmt(
+							_("html-renderer \"%s\" could not be executed."),
+							renderer.to_locale_string()));
+				} else {
+					v.get_statusline().show_error(strprintf::fmt(
+							_("html-renderer \"%s\" exited with error code %i."),
+							renderer.to_locale_string(), renderer_status));
+				}
+			}
```
Minoru, line 167: Do not convert `Filepath` into locale string; the logger needs UTF-8. `strprintf::fmt` already takes care of conversions, so just pass `renderer` and it'd work fine.

config_missing
Minoru, line 1: Looks like you meant to write tests for this, but didn't. Any particular problems you ran into that I can help with? Tests are important for getting this merged.

Minoru (commented):
Thank you! This is fairly raw and will need some further work, but it's a start.

dud-eisme:
does make fmt not format properly?

Minoru:
I see you're pushing commits, but haven't addressed all the comments yet. When you're done, please click the round button right of my avatar to re-request a review -- that way I'll know it's time to take another look.

Screenshot 2026-06-23 at 18-54-53 display an error if html-renderer failed to run by dud-eisme · Pull Request #3387 · newsboat_newsboat

Minoru:
> does make fmt not format properly?

It should, unless you have a different version of AStyle, in which case there might be bugs that lead to differences. Try our code-formatting-tools Dockerfile, or apply the diff that's shown at the end of the failing CI job.

dud-eisme:
ok the code formatting is passing now, and i have made the changes that you suggested

src/itemrenderer.cpp
Minoru: What I meant is: there is no need to print out the contents of `$?` and then parse it out of the output. `utils::run_program()` has a handle for the process it spawns, so it can obtain the exit code from that.

I now see we already ported that function to Rust. How familiar are you with Rust? Do you need help editing that part of the code?

3
Minoru: I believe this file, and all the test config files, can be deleted. Or am I wrong and they're required for something?

src/itemrenderer.cpp
dud-eisme: im pretty new to rust
but i should be able to modify the rust part too

dud-eisme:
> It should, unless you have a different version of AStyle, in which case there might be bugs that lead to differences. Try our code-formatting-tools Dockerfile, or apply the diff that's shown at the end of the failing CI job.

which version of AStyle is currently being used for this repo?
i have Artistic Style Version 3.6.16 currently installed on my machine

Minoru:
> which version of AStyle is currently being used for this repo?

3.1: https://github.com/newsboat/newsboat/blob/dbf2d3eae9a7652673e8d89e8a2eec0d942dec87/docker/code-formatting-tools.dockerfile

Minoru:
Force-pushes make it harder to review things, please avoid them. I can squash the commits when merging so only the final version of the code is preserved.

dud-eisme:
> Force-pushes make it harder to review things, please avoid them. I can squash the commits when merging so only the final version of the code is preserved.

oh ok i'll keep that in mind

dud-eisme:
hey @Minoru please review the commits

Minoru:
I haven't forgotten about this PR, but the earliest I'll get to it is Thursday.

src/feedretriever.cpp
```diff
@@ -284,13 +284,18 @@ rsspp::Feed FeedRetriever::download_filterplugin(const std::string& filter,
 			filter.c_str(),
 			nullptr
 		};
-	const std::string result = utils::run_program(argv, buf);
+	const nonstd::expected<std::string, int> result = utils::run_program(argv, buf);
+
+	if (!result) {
+		return rsspp::Feed();
+	}
```
Minoru, line 296: It'd be nice to log the exit code at `USERERROR` level so user can see the problem in their user-log.

src/formaction.cpp
```diff
@@ -875,7 +875,7 @@ std::string FormAction::bookmark(const std::string& url,
 			my_argv[1] = "-c";
 			my_argv[2] = cmdline.c_str();
 			my_argv[3] = nullptr;
-			return utils::run_program(my_argv, "");
+			return utils::run_program(my_argv, "").value();
```
Minoru: Calling `value()` without checking if there *is* a value will throw. Is the exception handled anywhere?

include/itemrenderer.h
```diff
@@ -7,6 +7,7 @@
 #include "dialog.h"
 #include "links.h"
 #include "textformatter.h"
+#include "../3rd-party/expected.hpp"
```
Minoru: Please see https://github.com/newsboat/newsboat/blob/master/doc/internal/code-style.markdown#order-of-includes 

Also, no need to use `../` here, include it like it's done in the other files.

src/itemrenderer.cpp
```diff
@@ -3,6 +3,7 @@
 #include <set>
 #include <sstream>
 
+#include "3rd-party/expected.hpp"
```
Minoru, line 6: https://github.com/newsboat/newsboat/blob/master/doc/internal/code-style.markdown#order-of-includes

src/itemrenderer.cpp
```diff
@@ -159,12 +160,15 @@ void render_html(
...
 		rnd.render(source, lines, thelinks, url);
+		return 0;
 	} else {
 		const char* argv[4];
 		argv[0] = "/bin/sh";
 		argv[1] = "-c";
 		const std::string renderer_locale_string = renderer.to_locale_string();
-		argv[2] = renderer_locale_string.c_str();
+
+		const std::string wrapped_command = renderer_locale_string;
```
Minoru: This variable is a leftover from intermediate versions of the code, and should be removed.

src/itemviewformaction.cpp
```diff
@@ -2,8 +2,10 @@
 
 #include <algorithm>
 #include <cstring>
+#include <iterator>
 #include <sstream>
 
+#include "3rd-party/expected.hpp"
```
Minoru, line 9: https://github.com/newsboat/newsboat/blob/master/doc/internal/code-style.markdown#order-of-includes

src/itemviewformaction.cpp
```diff
@@ -138,9 +140,39 @@ void ItemViewFormAction::prepare()
...
+								renderer.to_locale_string(), renderer_status));
+					}
+					std::tie(formatted_text, num_lines) =
+						item_renderer::source_to_stfl_list(
+							*item,
+							text_width,
+							window_width,
+							&rxman,
+							Dialog::Article);
+				}
```
Minoru: Hmm, not sure about this. Wouldn't it be better to fall back to the internal renderer? Do you have a rationale for this?

include/itemrenderer.h
dud-eisme: you want me to move that line to the lines with the <>(external) includes right?

Minoru: Hmm, I grepped for an example and it turns out that we have two approaches to including that header: we either put it onto a separate line, or we put it with the rest of Newsboat's internal headers.

Just take the `../` off then.

src/itemviewformaction.cpp
Minoru: Judging by the code, now it seems to show an error and display nothing. What's the idea behind that?

I'd prefer to discuss this before making any more changes. Perhaps I misunderstood some of the code.

include/itemrenderer.h
```diff
@@ -45,7 +46,7 @@ std::string to_plain_text(
 /// highlight the resulting text. \a links is filled with all the links
 /// found in the article (which is useful for article view, which lets
 /// users open the links by their number).
-std::pair<std::string, size_t> to_stfl_list(
+std::tuple<std::string, size_t, int> to_stfl_list(
```
Minoru: Wouldn't `std::expected<std::pair<std::string, size_t>, int>` be a better fit here? I'd clearly indicate that there is a "good" resulr (a pair) and an error (an int). Ideally the doc comment should explain what the int means :)

dud-eisme:
hey @Minoru , most probably you'd be busy, but pls review this pr :)

Minoru:
Yep, swamped with the PRs that need to be merged before the release. I don't have an estimate of when I'll get to this one, sorry.

dud-eisme:
no need to be sorry, haha
happens to all of us
i got term exams going on too ;-;

PR #3389, opened 2026-06-20
Handle all input in Newsboat itself

dennisschagt:
This is another step in getting rid of our STFL dependency (https://github.com/newsboat/newsboat/issues/232)
It is quite a big change so I definitely want to postpone merging until after the upcoming release (and I'm planning to test this for a while on my regular Newsboat setup)

Follow up work  which I prefer doing in a separate PR:
- ~~Handle input directly in Podboat as well without depending on STFL for input~~ -> https://github.com/newsboat/newsboat/pull/3420
- Move input handling from `FormAction::handle_qna_event` to `LineEdit` (at least the ones which are purely text based and do not require `History` access)

Some relevant parts of STFL which we no longer depend on after this PR:
- waiting for input: https://github.com/newsboat/stfl/blob/bbb2404580e845df2556560112c8aefa27494d66/base.c#L532-L537 -> implemented in our `FormAction`
- translating input to event string: https://github.com/newsboat/stfl/blob/bbb2404580e845df2556560112c8aefa27494d66/binding.c#L30-L84 -> implemented in our `FormAction`
- single-line input processing: https://github.com/newsboat/stfl/blob/bbb2404580e845df2556560112c8aefa27494d66/widgets/wt_input.c#L114-L175 -> Implemented on Rust side in `lineedit.rs` (`handle_event`, `insert_at_cursor`) and C++ side in `lineedit.cpp` (`LineEdit::handle_event`)

PR #3390, opened 2026-06-21
Reduce duplication in file and directory browsers

dennisschagt:
Relatively mechanical change by going through the diff between the file and directory browser formactions and adding a switch statement in cases of real differences (ignoring pure variable name differences)

This will cause merge conflicts with https://github.com/newsboat/newsboat/pull/3389 but I don't expect much difficulty in resolving them.
My preference is to merge this new PR before the other one.


dennisschagt:
Rebased to get results for required checks:
image

Minoru (approved):
Looks good, thank you!

I bumped FreeBSD in CI without looking at this first, so merge is blocked again. However, I don't think there will be any issues going from FreeBSD 15.0 to 15.1, so I'll just merge this as-is and keep an eye on the CI in master.

PR #3400, opened 2026-06-30
Prevent OPML import from injecting `exec:`/`filter:` URLs (RCE)

hyder365:
OPML import converts Liferea-style pipe-prefixed xmlUrl values (e.g.
`|~/script`) into `exec:` URLs, and honours the `filtercmd` attribute
by prepending `filter:`. Both schemes are executed via `sh -c` on the
next feed refresh -- so a shared OPML file (blogroll, etc.) can trigger
arbitrary command execution at the victim's privilege level when they
import it and reload feeds.

The fix drops both conversions during import:
- pipe-prefixed outlines are skipped (with a WARN log);
- `filtercmd` is ignored and the outline's own xmlUrl is imported as a
  regular feed.

Users who need program-generated or filtered feeds can still add
`exec:`/`filter:` entries to their urls file manually, which was
already the documented interface.

### DIFF
--- a/src/opml.cpp
+++ b/src/opml.cpp
@@ -109,37 +109,51 @@ void rec_find_rss_outlines(
 
 				std::string nurl = std::string(url);
 
-				// Liferea uses a pipe to signal feeds read from
-				// the output of a program in its OPMLs. Convert
-				// them to our syntax.
-				if (url.length() >= 1 && url[0] == '|') {
-					nurl = strprintf::fmt(
-							"exec:%s", url.substr(1));
-					LOG(Level::DEBUG,
-						"opml::import: liferea-style "
-						"url %s converted to %s",
-						url,
-						nurl);
-				}
+			// Liferea uses a pipe prefix in its OPMLs to signal feeds
+			// that are read from the output of a program. Converting
+			// such a URL into an `exec:` URL lets a shared OPML file
+			// run arbitrary shell commands on the victim's machine on
+			// the next feed refresh (the `exec:` scheme is passed
+			// verbatim to `sh -c`). Skip these entries; users who need
+			// a program-generated feed can add `exec:<script>` to
+			// their urls file manually.
+			bool skip_url = false;
+			if (url.length() >= 1 && url[0] == '|') {
+				LOG(Level::WARN,
+					"opml::import: skipping Liferea-style "
+					"pipe URL '%s' for security reasons; if "
+					"you trust it, add 'exec:%s' to your urls "
+					"file manually",
+					url,
+					url.substr(1));
+				skip_url = true;
+			}
 
-				// Handle OPML filters.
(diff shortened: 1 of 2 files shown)


Minoru:
Yep, that's an RCE for sure. I'd prefer a coordinated disclosure via security@newsboat.org, but oh well. I'll try to get a CVE for this, although I hear MITRE is hit and miss in regards of how responsive they are with allocations.

This PR fixes only one half of the problem, i.e. manual import using `--import-from-opml`. Newsboat also has an [OPML Online Subscription Mode](https://newsboat.org/releases/2.44/docs/newsboat.html#_opml_online_subscription_mode) which loads OPML from the Internet every time Newsboat is started. That last one is a bigger concern since the user can't realistically validate what's being loaded before it gets loaded. (We've meant to unify the code for them [for a long time](https://github.com/newsboat/newsboat/issues/292), but still haven't gotten around to it.)

UX-wise, logging at WARN level is useless because the user only sees it when they explicitly enable the log via command line. For manual import, these warnings should go to `stderr` and also to `Level::USERERROR`. For `opml-url`, I think just `USERERROR` is fine.

Are you willing to work on that, or should I take over? Either way, thank you for the discovery and the PR!

hyder365:
Feel free to take it from here

Minoru:
I've requested a CVE ID from MITRE.

Minoru:
MITRE still haven't assigned a CVE ID for this, and I'm done waiting. I'll have to publish this without an ID and add one retroactively.

(The main reason I waited was my broken computer which I failed to fix and now replaced with a new one.)

Minoru:
I had to force-push in order to resolve merge conflicts with the master branch. I want to keep this PR's history linear in case someone needs to backport patches onto an older version.

The only changes I had to make during rebase is re-run the `make fmt` to account for the changes done to src/opml.cpp on the master branch. Here are screenshots of the output of `git range-diff origin/master ee059961b381504c1b32bd9936223cd41b9a286f bugfix/hyder365-opml-inject-rce`:

Screenshot From 2026-09-06 18-11-23
(continued as it didn't fit on a single screen)
Screenshot From 2026-09-06 18-11-39

Also, still no CVE ID from MITRE.

Minoru:
One more rebase to pull in a fix for FreeBSD build (https://github.com/newsboat/newsboat/pull/3440). No actual changes:

```
$ git range-diff origin/master 0cf205020 bugfix/hyder365-opml-inject-rce
-:  --------- > 1:  a68f70bb8 Prevent OPML import from injecting `exec:`/`filter:` URLs (RCE)
-:  --------- > 2:  fad693989 Fix formatting
-:  --------- > 3:  e49a8d9fc Log to stderr and user error log
-:  --------- > 4:  83259a6ec Fix clang-tidy warning
```

Issue #3416, opened 2026-07-20
Provide `date`/`feeddate` attribtues with fixed format


Minoru:
#3395 uncovered that current formats are hard to use, and impossible to use portably since they depend on the locale. As suggested in https://github.com/newsboat/newsboat/commit/44f8ceed2371de42bac533733bdd18dab4f41c67, we should provide a different set of attributes whose format is fixed and can be matched against more easily.

That commit suggests RFC 3339 format (e.g. 1985-04-12T23:20:50.52Z). While readable, it's hard to match against. Perhaps we could provide a new set of operators for matching against time strings, but that feels like a chunk of work.

Alternatively, the new attributes could use Unix epoch, which is just a number. We already have operators to match against numbers, so it'd fit like a glove. As for readability, our config format supports backticks, so it's possible to write something like ``` date > `date +%s -d'2000-01-01'` and date < `date +%s -d'3000-01-01'` ```. Not great, not terrible. Also, it won't work when typing the filter interactively from within Newsboat, and probably when viewing filters in Newsboat (the user will see results of evaluation of backticks, not the commands).

PR #3417, opened 2026-07-20
Document `date`/`feeddate` filter attribute format

### LINKED ISSUE #3395: how to use date filter?
I have some feeds (eg nitter) where sometimes deleted articles reappear on refresh.
I thought to use age but sometimes servers break for long time so i dont want to rely on age. 
I wanted to look at using pubdate filter but the docs are not clear or gave any examples on how to use it.

I tried using stuff like
"date>= \"12-05-2026\""
 Or
"date<= \"12-05-2026\""

But it's not showing any articles after that.

### LINKED ISSUE #3399: CI: Issues with required checks not being satisfied due to path filters
It looks like we are running into a longstanding Github Actions issue/limitation.
When a pipeline is skipped due to a path filter, it does not report an outcome and thus does not satisfy the related required checks (see for example: https://github.com/newsboat/newsboat/pull/3390)
The Github docs mention this as an issue but do not provide a proper alternative. From [Github docs on "handling skipped but required checks"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks):
>  If a workflow is skipped due to path filtering, branch filtering or a commit message, then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging.
>
> If, however, a job within a workflow is skipped due to a conditional, it will report its status as "Success". For more information, see Using conditions to control job execution.

I guess we can use conditional checks at the job/step level instead (as mentioned in the docs) but it looks like there is no built-in way to do that (so might need a dependency on a 3rd-party action like [dorny/paths-filter](https://github.com/dorny/paths-filter))

To be honest, I'm quite disappointed with Github's lack of reaction on community issues with significant upvotes which show this has been an annoying limitation for a while now, e.g.:
- https://github.com/orgs/community/discussions/44490
- https://github.com/orgs/community/discussions/13690

Minoru:
This was contributed by der-lyse. The the first commit for the thorough explanation.

I reformatted the text a tiny bit and added comments for translators to avoid making the situation worse.

We should add attributes that don't have those problems: https://github.com/newsboat/newsboat/issues/3416

Fixes #3395.


include/rssfeed.h
```diff
@@ -67,6 +67,9 @@ class RssFeed : public Matchable {
 
 	std::string pubDate() const
 	{
+		// i18n: do not translate this string. It was a mistake to make it
+		// translatable, but undoing it might break some users' filters.
+		// See 2c56225b1cf5ee836d4aa4f875b0d591911811ec.
```
dennisschagt: It looks like this commit does not currently exist (I guess it got lost in a rebase)

src/rssitem.cpp
```diff
@@ -122,6 +122,9 @@ void RssItem::set_unread(bool u)
 
 std::string RssItem::pubDate() const
 {
+	// i18n: do not translate this string. It was a mistake to make it
+	// translatable, but undoing it might break some users' filters.
+	// See 2c56225b1cf5ee836d4aa4f875b0d591911811ec.
```
dennisschagt: Same here

dennisschagt (approved):
Nice work, LGTM

Minoru:
You're right, I rebased it without thinking about those ids. I force-pushed a fix, here's a range-diff:

image

dennisschagt:
@Minoru I believe the wrong issue got closed: https://github.com/newsboat/newsboat/issues/3399
Did you want to link/close this one instead or am I possibly missing something? https://github.com/newsboat/newsboat/issues/3395

Minoru:
You're right, I mis-typed the issue number in the PR description. I now re-opened/closed everything to fix it. Thank you for being so vigilant :)

Issue #3345, opened 2026-05-24
Simplify FileUrlReader


dennisschagt:
From https://github.com/newsboat/newsboat/pull/3339#issuecomment-4452643839
> apparently Newsboat writes to urls file only when importing from OPML. In that scenario it always exits right after writing, so there is really no point in updating the internal state anymore. I think it's better to turn FileUrlReader::write_config into static function, or maybe even a standalone one, which just takes the FileUrlReader, calls get_urls()/get_tags() as appropriate, and writes to the specified file. Actually, UrlReader::add_url() also exists just for OPML import, so perhaps the entire import should be refactored to better separate what OPML import needs from what Newsboat TUI needs. (That'd be for a separate PR, this one is large enough already.)

PR #3423, opened 2026-07-30
UrlReader: Combine `urls` and `tags` into single storage

dennisschagt:
Avoid urls/tags lists getting out of sync.
Preparation for https://github.com/newsboat/newsboat/issues/3345.

### DIFF
--- a/include/urlreader.h
+++ b/include/urlreader.h
@@ -1,7 +1,6 @@
 #ifndef NEWSBOAT_URLREADER_H_
 #define NEWSBOAT_URLREADER_H_
 
-#include <map>
 #include <optional>
 #include <string>
 #include <vector>
@@ -15,6 +14,12 @@ namespace newsboat {
 /// \brief Base class for classes that supply Newsboat with feed URLs.
 class UrlReader {
 public:
+	struct FeedUrl {
+		std::string url;
+		FeedOrigin origin;
+		std::vector<std::string> tags;
+	};
+
 	UrlReader() = default;
 	virtual ~UrlReader() = default;
 
@@ -32,7 +37,7 @@ public:
 	virtual std::string get_source() const = 0;
 
 	/// \brief A list of feed URLs with their origin
-	const std::vector<std::pair<std::string, FeedOrigin>>& get_urls() const;
+	const std::vector<UrlReader::FeedUrl>& get_urls() const;
 
 	/// \brief Tags of feed that has url `url`.
 	std::vector<std::string>& get_tags(const std::string& url);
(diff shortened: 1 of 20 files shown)


include/urlreader.h
```diff
@@ -32,7 +37,7 @@ class UrlReader {
 	virtual std::string get_source() const = 0;
 
 	/// \brief A list of feed URLs with their origin
-	const std::vector<std::pair<std::string, FeedOrigin>>& get_urls() const;
+	const std::vector<UrlReader::FeedUrl>& get_urls() const;
```
Minoru: Nit: the doc comment should be updated to mention that the tags are there too. (I'd make a code suggestion, but for some reason GitHub won't let me.)

src/controller.cpp
```diff
@@ -829,11 +828,11 @@ void Controller::reload_urls_file()
...
 
-	for (const auto& [url, origin] : urlcfg->get_urls()) {
-		const auto feed = feedcontainer.get_feed_by_url(url);
+	for (const auto& feed_url : urlcfg->get_urls()) {
+		const auto feed = feedcontainer.get_feed_by_url(feed_url.url);
 		if (feed) {
-			feed->set_origin(origin);
-			feed->set_tags(urlcfg->get_tags(url));
+			feed->set_origin(feed_url.origin);
+			feed->set_tags(urlcfg->get_tags(feed_url.url));
```
Minoru: I think the tags are already present in `feed_url`:

```suggestion
			feed->set_tags(feed_url.tags);
```

src/urlreader.cpp
```diff
@@ -7,21 +7,32 @@
...
+	auto it = std::find_if(feed_urls.begin(),
+	feed_urls.end(), [&url](const FeedUrl& feed_url) {
+		return feed_url.url == url;
+	});
+	if (it != feed_urls.end()) {
+		return it->tags;
+	} else {
+		return dummy;
+	}
 }
```
Minoru: Is this method needed at all? I didn't check thoroughly, but it looks like this PR already converted all callsites of `get_tags()` to `get_urls()`. If that's not the case and you don't already have a fix for this TODO in the pipeline, please open an issue about it, preferably with suggestions on what exactly new UrlReader interface should accomplish.

Minoru (approved):
Thanks!

src/controller.cpp
dennisschagt: Indeed, suggestion applied (manually, combined with other related changes)

src/urlreader.cpp
dennisschagt: There are still two places left which needed to get the tags given a url (in `src/{opml.cpp,controller.cpp}`)
I've changed the UrlReader interface slightly to instead return a pointer with a nullptr option (alternatively, I could have used `std::optional` with a reference but that would have required a somewhat ugly `std::reference_wrapper`)

dennisschagt: I'm merging it already but if you prefer a different interface, just let me know and I will take it along in a followup PR

PR #3426, opened 2026-08-18
Reject special URLs from online OPML

acts-1631:
Online OPML subscription mode accepts xmlUrl attributes without validating their scheme. A server controlling the configured opml-url can return exec: or filter: entries, which are later passed to a shell during feed refresh. This allows command execution as the Newsboat user without modifying the local urls file.

Reject non-HTTP(S) xmlUrl values in OpmlUrlReader and report them as user-visible errors. This keeps exec: and filter: available in explicitly local configuration while preventing remote OPML from activating them.

This is a follow-up to #3400. That PR fixes manual --import-from-opml conversions of Liferea pipe URLs and filtercmd; this change covers the separate online OPML subscription path and does not duplicate those changes.

### DIFF
--- a/src/opmlurlreader.cpp
+++ b/src/opmlurlreader.cpp
@@ -68,6 +68,16 @@ void OpmlUrlReader::handle_node(xmlNode* node, const std::string& tag)
 			(char*)xmlGetProp(node, (const xmlChar*)"xmlUrl");
 		if (rssurl && strlen(rssurl) > 0) {
 			std::string theurl(rssurl);
+			xmlFree(rssurl);
+
+			if (!utils::is_http_url(theurl)) {
+				LOG(Level::USERERROR,
+					"OpmlUrlReader: skipping non-HTTP(S) URL from "
+					"remote OPML: %s",
+					theurl);
+				return;
+			}
+
 			urls.push_back({theurl, FeedOrigin{}});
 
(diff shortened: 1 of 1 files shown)


Minoru:
Thanks! This looks like a good start, but needs a test. Also, the `xmlFree` change is a bit inelegant; I'd prefer to convert the string upfront so we don't have to call `xmlFree` way further from where the pointer was allocated. This PR also needs a rebase on top of current master to fix the merge conflicts.

I'll do it all myself to avoid the code review back-and-forth; I'll like to get those fixes out now.

Minoru:
I ended up rewriting this entirely, so I'll just build on top of the existing PR.

Minoru:
Turns out I misunderstood this PR. Initially I thought it fixes the same problem as #3400, but for `opml-url`. Now I see it actually fixes a different problem: remote OPML source could contain `exec:` and `filter:` URLs which would be copied to Newsboat's urls file and execute binaries on the user's machine. This bug actually exists in #3400 too, and should have been fixed there as well.

Despite my claims in https://github.com/newsboat/newsboat/pull/3400#issuecomment-5230429131, the #3400 bug doesn't exist in `opml-urls` because `opml-urls` doesn't process `|` and `filtercmd`.

PR #3427, opened 2026-08-18
Reject control characters in podcast queue data

acts-1631:
Podcast enclosure URLs are written to Podboat's line-oriented queue without rejecting control characters. A remote feed can encode a newline in an enclosure URL, causing Newsboat to write additional queue records. Podboat then interprets attacker-controlled fields as a download URL and output pathname.

Reject control characters in enclosure URLs and generated filenames before writing queue records. Podboat also ignores malformed queue entries and fields containing control characters when reloading the queue.

This prevents remote podcast metadata from injecting queue records and redirecting downloads to attacker-selected paths.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -731,6 +731,10 @@ void Controller::replace_feed(RssFeed& oldfeed, RssFeed& newfeed, unsigned int p
 			v->get_statusline().show_error(
 				strprintf::fmt(_("Failed to open queue file: %s."), result.extra_filename));
 			break;
+		case EnqueueStatus::INVALID_ENQUEUE_DATA:
+			v->get_statusline().show_error(
+				_("Item has a non-HTTP enclosure URL."));
+			break;
 		}
 	}
 
--- a/src/itemutils.cpp
+++ b/src/itemutils.cpp
@@ -38,6 +38,10 @@ bool enqueue_item_enclosure(RssItem& item, RssFeed& feed,
 			v.get_statusline().show_error(
 				strprintf::fmt(_("Failed to open queue file: %s."), result.extra_filename));
 			return false;
+		case EnqueueStatus::INVALID_ENQUEUE_DATA:
+			v.get_statusline().show_error(
+				_("Item has a non-HTTP enclosure URL."));
+			return false;
 		}
 
 		// Not reachable, all switch cases return a result already,
--- a/src/queueloader.cpp
+++ b/src/queueloader.cpp
@@ -121,9 +121,6 @@ void QueueLoader::update_from_queue_file(CategorizedDownloads& downloads) const
 			continue;
 		}
 
-		LOG(Level::DEBUG,
-			"QueueLoader::reload: loaded `%s' from queue file",
-			line);
 		const std::vector<std::string> fields = utils::tokenize_quoted(line);
 		bool url_found = false;
 
@@ -142,6 +139,27 @@ void QueueLoader::update_from_queue_file(CategorizedDownloads& downloads) const
 			}
 			continue;
 		}
+
+		const bool valid_status = fields.size() < 3
+			|| fields[2] == "missing"
+			|| fields[2] == "downloaded"
+			|| fields[2] == "played"
+			|| fields[2] == "finished";
+		// Control characters are preserved inside quoted fields.
+		const bool valid_fields = !utils::contains_control_characters(fields[0])
+			&& fields.size() <= 3
+			&& valid_status
+			&& (fields.size() < 2
+				|| !utils::contains_control_characters(fields[1]));
+		if (!valid_fields) {
+			LOG(Level::USERERROR,
+				"QueueLoader: skipping malformed queue entry");
+			continue;
+		}
+
+		LOG(Level::DEBUG,
+			"QueueLoader::reload: loaded `%s' from queue file",
+			line);
 
 		for (const auto& dl : downloads.to_keep) {
 			if (fields[0] == dl.url()) {
--- a/src/queuemanager.cpp
+++ b/src/queuemanager.cpp
@@ -17,7 +17,18 @@ QueueManager::QueueManager(ConfigContainer* cfg_, Filepath queue_file)
 
 EnqueueResult QueueManager::enqueue_url(RssItem& item, RssFeed& feed)
 {
-	const std::string& url = item.enclosure_url();
+	const std::string url = utils::sanitize_url(item.enclosure_url());
+	if (!utils::is_http_url(url)) {
+		LOG(Level::USERERROR,
+			"QueueManager: refusing to enqueue a non-HTTP podcast URL `%s' "
+			"from feed `%s'",
+			url,
+			feed.rssurl());
+		EnqueueResult result;
+		result.status = EnqueueStatus::INVALID_ENQUEUE_DATA;
+		return result;
+	}
+
 	const Filepath filename = generate_enqueue_filename(item, feed);
 
 	std::fstream f;
@@ -75,7 +86,7 @@ std::string get_hostname_from_url(const std::string& url)
 
 Filepath QueueManager::generate_enqueue_filename(RssItem& item, RssFeed& feed)
 {
-	const std::string& url = item.enclosure_url();
+	const std::string url = utils::sanitize_url(item.enclosure_url());
 	const std::string& title = utils::utf8_to_locale(item.title());
 	const time_t pubDate = item.pubDate_timestamp();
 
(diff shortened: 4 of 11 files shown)


src/queuemanager.cpp
```diff
@@ -18,7 +18,23 @@ QueueManager::QueueManager(ConfigContainer* cfg_, Filepath queue_file)
 EnqueueResult QueueManager::enqueue_url(RssItem& item, RssFeed& feed)
 {
 	const std::string& url = item.enclosure_url();
+	if (!utils::is_http_url(url) || utils::contains_control_characters(url)) {
+		LOG(Level::USERERROR,
+			"QueueManager: refusing to enqueue an invalid podcast URL");
+		EnqueueResult result;
+		result.status = EnqueueStatus::INVALID_ENQUEUE_DATA;
+		return result;
+	}
```
Minoru: rust/libnewsboat already depends on https://crates.io/crates/percent-encoding. How about sanitizing the URL instead of dropping it?

src/queuemanager.cpp
```diff
@@ -18,7 +18,23 @@ QueueManager::QueueManager(ConfigContainer* cfg_, Filepath queue_file)
...
+
 	const Filepath filename = generate_enqueue_filename(item, feed);
+	if (utils::contains_control_characters(filename.to_locale_string())) {
+		LOG(Level::USERERROR,
+			"QueueManager: refusing to enqueue a filename with control "
+			"characters");
+		EnqueueResult result;
+		result.status = EnqueueStatus::INVALID_ENQUEUE_DATA;
+		return result;
+	}
```
Minoru: This is a fairly heavy-handed solution, preventing the user from listening to podcast just because there is a wrong byte in the filename. Wrong byte doesn't necessarily mean malice.

Let's replace control characters with an underscore instead. Newsboat already does that for article filenames when they're saved to disk (cf. `restrict-filename` setting), so that should be a good compromise.

Minoru (commented):
Thank you! Given my experience with MITRE in https://github.com/newsboat/newsboat/pull/3400#issuecomment-5512435296, I won't request a CVE ID for this right away, but I'll definitely highlight this as a security fix and release this along with others when they're ready.

Once you've addressed my comments below, please add tests for the `QueueManager` and `QueueLoader` to demonstrate new behaviour.

acts-1631:
Ok, have a look at this one if you will.

src/queuemanager.cpp
```diff
@@ -73,9 +83,9 @@ std::string get_hostname_from_url(const std::string& url)
 	return hostname;
 }
 
-Filepath QueueManager::generate_enqueue_filename(RssItem& item, RssFeed& feed)
+Filepath QueueManager::generate_enqueue_filename(RssItem& item, RssFeed& feed,
+	const std::string& url)
 {
-	const std::string& url = item.enclosure_url();
```
Minoru, line 78: I think it'd be safer to sanitize the URL in `generate_enqueue_filename()` instead of passing it in. Currently there's nothing preventing someone from adding another call to this method and passing the wrong URL.

src/queuemanager.cpp
```diff
@@ -128,6 +138,12 @@ EnqueueResult QueueManager::autoenqueue(RssFeed& feed)
 
 		const auto enclosure_type = item->enclosure_type();
 		const auto enclosure_url = item->enclosure_url();
+		if (!utils::is_http_url(enclosure_url)) {
+			LOG(Level::DEBUG,
+				"QueueManager::autoenqueue: skipping enclosure with "
+				"invalid URL");
+			continue;
+		}
```
Minoru, line 149:
```suggestion
		if (!utils::is_http_url(enclosure_url)) {
			LOG(Level::DEBUG,
				"QueueManager::autoenqueue: skipping enclosure with "
				"invalid URL `%s'", enclosure_url);
			continue;
		}
```

src/queuemanager.cpp
```diff
@@ -17,8 +17,18 @@ QueueManager::QueueManager(ConfigContainer* cfg_, Filepath queue_file)
...
-	const std::string& url = item.enclosure_url();
-	const Filepath filename = generate_enqueue_filename(item, feed);
+	const std::string url = utils::sanitize_url(item.enclosure_url());
+	if (!utils::is_http_url(url)) {
+		LOG(Level::USERERROR,
+			"QueueManager: refusing to enqueue an invalid podcast URL");
+		EnqueueResult result;
+		result.status = EnqueueStatus::INVALID_ENQUEUE_DATA;
+		return result;
+	}
```
Minoru, line 30: This error message lacks detail, so let's add some:

```suggestion
	const std::string url = utils::sanitize_url(item.enclosure_url());
	if (!utils::is_http_url(url)) {
		LOG(Level::USERERROR,
			"QueueManager: refusing to enqueue an non-HTTP podcast URL `'%s' from feed `%s'",
			url, feed.rssurl());
		EnqueueResult result;
		result.status = EnqueueStatus::INVALID_ENQUEUE_DATA;
		return result;
	}
```

src/itemutils.cpp
```diff
@@ -38,6 +38,10 @@ bool enqueue_item_enclosure(RssItem& item, RssFeed& feed,
 			v.get_statusline().show_error(
 				strprintf::fmt(_("Failed to open queue file: %s."), result.extra_filename));
 			return false;
+		case EnqueueStatus::INVALID_ENQUEUE_DATA:
+			v.get_statusline().show_error(
+				_("Item has an invalid enclosure URL."));
```
Minoru:
```suggestion
				_("Item has a non-HTTP enclosure URL."));
```

src/controller.cpp
```diff
@@ -733,6 +731,9 @@ void Controller::replace_feed(RssFeed& oldfeed, RssFeed& newfeed, unsigned int p
 			v->get_statusline().show_error(
 				strprintf::fmt(_("Failed to open queue file: %s."), result.extra_filename));
 			break;
+		case EnqueueStatus::INVALID_ENQUEUE_DATA:
+			// Invalid remote metadata is skipped by QueueManager.
+			break;
```
Minoru, line 737: This should display an error similar to how it's done in the itemutils.

src/queueloader.cpp
```diff
@@ -143,6 +140,26 @@ void QueueLoader::update_from_queue_file(CategorizedDownloads& downloads) const
...
+		const bool valid_status = fields.size() < 3
+			|| fields[2] == "missing"
+			|| fields[2] == "downloaded"
+			|| fields[2] == "played"
+			|| fields[2] == "finished";
+		const bool valid_fields = !utils::contains_control_characters(fields[0])
+			&& fields.size() <= 3
+			&& valid_status
+			&& (fields.size() < 2
+				|| !utils::contains_control_characters(fields[1]));
```
Minoru, line 153: Why are these checks necessary? `utils::tokenize_quoted()` splits on " \n\r\t", so none of those will be present in the fields.

test/queueloader.cpp
```diff
@@ -677,6 +679,35 @@ TEST_CASE("reload() skips empty lines in the queue file", "[QueueLoader]")
...
+TEST_CASE("reload() skips queue entries containing control characters",
+	"[QueueLoader]")
+{
+	test_helpers::TempFile queueFile;
+	std::ofstream queue(queueFile.get_path().to_locale_string());
+	queue << R"("https://example.com/bad\nurl" "/tmp/bad.mp3")" << '\n';
+	queue << R"(https://example.com/bad-filename.mp3 "/tmp/bad\nfilename.mp3")" << '\n';
+	queue << R"(https://example.com/too-many-fields.mp3 "/tmp/bad.mp3" missing extra)" << '\n';
+	queue << R"(https://example.com/unknown-status.mp3 "/tmp/bad.mp3" unknown)" << '\n';
+	queue << R"(https://example.com/good.mp3 "/tmp/good.mp3")" << '\n';
```
Minoru: `\r` and `\t` should be tested as well. `\n` is handled by `getline()`, so this test doesn't actually exercise the checks further inside the method.

Minoru (changes requested):
I expected this review cycle to be about sweating the small details, but upon closer inspection I found a couple more issues to fix. Please see the comments below.

I assume this is generated by an LLM? Are your reading the code it produces before submitting it for review?

acts-1631:
Hi and thanks for reviewing. The follow-up moves URL sanitization into generate_enqueue_filename(), sanitizes the final filename, improves the error messages, updates the auto-enqueue logging, and fixes the formatting issues. I also updated the QueueLoader test to use actual CR and TAB characters inside quoted fields. (I left those checks in place because quoted fields preserve control characters, and the updated test now covers that case explicitly.)

src/queueloader.cpp
Minoru: >  I also updated the QueueLoader test to use actual CR and TAB characters inside quoted fields. (I left those checks in place because quoted fields preserve control characters, and the updated test now covers that case explicitly.)

Minoru (approved):

PR #3428, opened 2026-08-18
Bound RSS response buffering

acts-1631:
Newsboat accumulates each decoded RSS/Atom response in a std::string without a size limit. A remote feed server can stream an oversized or compressed response and exhaust client memory before parsing begins.

Add an optional limit to CurlDataReceiver and apply a 50 MiB decoded-response cap to RSS/Atom parsing. The write callback aborts the transfer before appending data that would exceed the cap. Other API consumers retain their existing behavior.

The initial 50 MiB limit is a conservative choice and can be adjusted if maintainers prefer a different value.

### DIFF
--- a/include/curldatareceiver.h
+++ b/include/curldatareceiver.h
@@ -1,6 +1,7 @@
 #ifndef NEWSBOAT_CURLDATARECEIVER_H_
 #define NEWSBOAT_CURLDATARECEIVER_H_
 
+#include <cstddef>
 #include <memory>
 #include <string>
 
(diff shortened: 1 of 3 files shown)


Minoru:
I'm postponing this one until the next release. This *is* a vulnerability, but the worst it can do is make Newsboat unavailable (crash it).

A few questions:

* how did you test this?
* why didn't you apply the same limit to remote APIs?
* why didn't you apply it to filters?
* how does this work with retry logic in `FeedRetriever`? Wouldn't it treat it as failure and retry the download?

Also:

* I feel like the user should be notified if a feed is too large to be downloaded
* I'd make the limit configurable, because there is bound to be a user that parses giant feeds

acts-1631:
Thanks for the review. I pushed an updated version in 496ce3f7.

The response limit now applies to direct RSS/Atom downloads, remote API responses, and downloads performed through `utils::retrieve_url()`, including the input to `filter:` URLs. The limit is configurable through `download-max-size`, with a default of 50 MiB. Setting it to 0 disables the limit.

Oversized RSS responses now produce a specific error mentioning the configured limit. They are not retried: the write callback aborts the transfer, libcurl reports `CURLE_WRITE_ERROR`, and `Parser::parse_url()` throws before `FeedRetriever` reaches its retry loop.

PR #3431, opened 2026-08-29
ci: use job-level conditional instead of path filter for no_dbg_macros

chrisdebian:
Fixes the pattern described in #3399 for one workflow, as a small, reviewable first step.

GitHub Actions leaves a required check permanently "Pending" when its workflow is skipped by a trigger-level `paths:` filter — as you noted, this blocks merges even when nothing relevant changed. Per GitHub's own docs (quoted in #3399), a *job* skipped via an `if:` conditional reports "Success" instead, which satisfies required-check rules.

This PR moves the path filter for `no_dbg_macros.yml` out of the trigger (`on: push/pull_request: paths:`) and into a `changes` job using `dorny/paths-filter@v4`, gating the real job on that job's output.

**Update**: per @Minoru's comment, this now also applies the same change to `asciidoctor.yml` and `i18nspector.yml` — the only other two workflows using the same trigger-level `paths:` pattern.

### DIFF
--- a/.github/workflows/no_dbg_macros.yml
+++ b/.github/workflows/no_dbg_macros.yml
@@ -2,15 +2,31 @@ name: "No dbg! macros in Rust code"
 
 on:
   push:
-    paths: &included_paths
-      - '**.rs'
-      - '.github/**'
   pull_request:
-    paths: *included_paths
 
 jobs:
+  changes:
+    name: "Detect relevant changes"
+    runs-on: ubuntu-24.04
+    outputs:
+      relevant: ${{ steps.filter.outputs.relevant }}
+    steps:
+      - name: Check out the code
+        uses: actions/checkout@v7
+
+      - name: Filter paths
+        uses: dorny/paths-filter@v4
+        id: filter
+        with:
+          filters: |
+            relevant:
+              - '**.rs'
+              - '.github/**'
+
   check_for_dbg_macros:
     name: "No dbg! macros in Rust code"
+    needs: changes
+    if: needs.changes.outputs.relevant == 'true'
     runs-on: ubuntu-24.04
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/no_dbg_macros.yml
```diff
@@ -2,15 +2,31 @@ name: "No dbg! macros in Rust code"
...
 
 jobs:
+  changes:
+    name: "Detect relevant changes"
+    runs-on: ubuntu-24.04
+    outputs:
+      relevant: ${{ steps.filter.outputs.relevant }}
+    steps:
+      - name: Check out the code
+        uses: actions/checkout@v7
```
dennisschagt, line 15: I believe this checkout is only necessary because we run these checks "on push".
For the "on pull_request" case, I believe `dorny/paths-filter` can run without a checkout..

@Minoru Is the on `push` trigger meant just for  merge commits on master or do you also actively use/check these builds before creating a PR?
(I, and I assume most other contributors, only push to my own clone of Newsboat and create my PRs directly from there)

Anyway, I think these changes might actually increase CI time on average for this specific workflow (as it adds a checkout while it only guards a checkout+grep below)
However, I do think the general idea looks good so this might be nice to apply to the workflows with a longer build time (where there is more CI build time to save)

Before doing so, it might be nice to wait a bit to see if @Minoru agrees.

Minoru: > @Minoru Is the on push trigger meant just for merge commits on master or do you also actively use/check these builds before creating a PR?

It's for both. Treating this repo as my personal repo makes it a bit easier to make releases since I don't have to juggle two Git remotes, but if that blocks some improvements, I'll make a proper GitHub fork and work off that.

> However, I do think the general idea looks good so this might be nice to apply to the workflows with a longer build time (where there is more CI build time to save)

I agree. Let's do this for all the workflows.

chrisdebian:
Done — pushed the same fix to `asciidoctor.yml` and `i18nspector.yml`. All three workflows using this trigger-level `paths:` pattern now use the job-level conditional instead.

PR #3440, opened 2026-09-12
Relax test for CURL constants

Minoru:
curl commit a55731050e8c3dbea0b96205cf916443118f6acb[1] changed the value of `CURLAUTH_ANY`. The change was released in curl 8.22, but our Rust code uses curl-sys that matches 8.21. As a result, Rust's idea about the value of `CURLAUTH_ANY` didn't match that of C++, and a C++ test failed.

This commit relaxes the checking somewhat. The goal is to ensure that Rust returns "sane" value, without requiring strict equality.

I didn't convert checks for other constants because they verify that Newsboat abides by user's settings.

1. https://github.com/curl/curl/commit/a55731050e8c3dbea0b96205cf916443118f6acb

### DIFF
--- a/test/utils.cpp
+++ b/test/utils.cpp
@@ -1631,7 +1631,24 @@ TEST_CASE(
 	// (via https://github.com/curl/curl/pull/20416)
 	// To avoid mismatches between the curl version bundled on the rust side (curl-sys) and the
 	// system curl version, apply manual casting to uint32_t.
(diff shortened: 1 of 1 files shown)


Minoru:
This problem breaks CI for all current PRs, so it's somewhat urgent to fix this. I intend to merge it once I'm ready to merge #3400, because that one blocks work on updating the translations.

dennisschagt (approved):

Minoru:
Thanks for the quick review! I ended up making a bunch more changes, but they're all in line with the original logic, so I think I'll just merge this already.

Issue #3408, opened 2026-07-12
Replace json-c uses with 3rd-party/json.hpp


dennisschagt:
From https://github.com/newsboat/newsboat/issues/3404:
> Since it's the second or third time json-c brakes our build for no good reason, move to json.hpp. (Big task, I intended to punt it and then drop it when everything migrates to Rust)

Some of our remote API implementations already use json.hpp, but there are still several left:
```
> git grep '#include .*json.h' | grep -v 3rd-party
include/newsblurapi.h:#include <json.h>
include/ocnewsapi.h:#include <json-c/json.h>
src/feedhqapi.cpp:#include <json.h>
src/freshrssapi.cpp:#include <json.h>
src/inoreaderapi.cpp:#include <json.h>
src/ocnewsapi.cpp:#include <json-c/json.h>
src/oldreaderapi.cpp:#include <json.h>
```

- [ ] newsblurapi
- [x] ocnewsapi -> https://github.com/newsboat/newsboat/pull/3414
- [ ] feedhqapi
- [ ] freshrssapi
- [ ] inoreaderapi
- [ ] oldreaderapi

arthurhen00:
Hi! I'd like to work on the feedhqapi part of this issue. I'm currently replacing its json-c usage with nlohmann/json and will submit a PR once the changes are tested.

PR #3442, opened 2026-09-13
Inoreader API: Use nlohmann json instead of json-c

arthurhen00:
Part of #3408.

Replace the remaining json-c usage in the Inoreader API with the nlohmann/json library.

The implementation follows the same approach as #3436, using json::parse() and handling nlohmann/json exceptions for invalid or unexpected responses.

This also removes the json-c version-specific handling around category iteration, since nlohmann/json allows iterating over the categories directly.


Minoru:
Hey, just wanted to let you know that I'm seeing the PRs, but haven't gotten to them yet because I need to deal with security-related ones first.

One question off the bat: did you test those changes? Cover letters don't seem to mention this at all.

Also, the next release is in two weeks, so depending on how confident I feel about those changes I might postpone merging them until after the release. That'd give us 3 more months to test everything before it's released.

arthurhen00:
> Hey, just wanted to let you know that I'm seeing the PRs, but haven't gotten to them yet because I need to deal with security-related ones first.
> 
> One question off the bat: did you test those changes? Cover letters don't seem to mention this at all.
> 
> Also, the next release is in two weeks, so depending on how confident I feel about those changes I might postpone merging them until after the release. That'd give us 3 more months to test everything before it's released.


Hi, I tested all three changes locally.

For FeedHQ (#3436), FreshRSS (#3437), and Inoreader (#3442) I added temporary local tests around the actual `get_subscribed_urls()` implementation using the `HttpTestServer`.

For FeedHQ, I covered:

- parsing a valid subscription;
- the generated feed URL and tags;
- subscriptions without an id;
- subscriptions without a title;
- invalid JSON.

I ran the exact same tests against both the old json-c implementation and the new nlohmann/json implementation.

For FreshRSS, I also covered categories and null category labels:

- valid subscription with categories;
- missing id;
- missing title;
- category with a null label;
- invalid JSON.

Again, I ran the same tests against both implementations.

For Inoreader I tested the same kinds of cases, also through `get_subscribed_urls()` and `HttpTestServer`. There was one extra complication: the Inoreader subscription-list URL is hard-coded in `inoreaderapi.cpp`, so it can't normally be pointed at the local HTTP test server through configuration.

For the local test only, I added a temporary URL override so that the request made by `get_subscribed_urls()` would go to the local `HttpTestServer` instead of `inoreader.com`. I applied the exact same temporary override to both the old json-c version and the new nlohmann/json version, so both implementations were exercised in the same way.

These temporary tests were only used locally and aren't part of the PRs at the moment.

I also ran the full test suite for each of the three PR branches.

PR #3445, opened 2026-09-19
OPML: only import URLs with supported schemas

Minoru:
The problem was reported, and a fix suggested, in
https://github.com/newsboat/newsboat/pull/3426.


Minoru:
Force-push to fix little errors in OPML files used by tests: I forgot to update their titles, and `query:` example used invalid regex.

PR #3448, opened 2026-09-20
Update POT and PO files

Minoru:


Minoru:
@newsboat/translators, please update your respective PO files by October 4th (two weeks from now). The release is delayed because of important PRs that arrived late in the release cycle.

(In case you don't want to be pinged about Newsboat translations anymore, drop a comment here or email me.)

